PR #5, opened 2017-05-02
Change debug print to hex

AndreasBomholtz:
I change the debug print to show te hex bytes instead of the string.

### DIFF
--- a/bellows/uart.py
+++ b/bellows/uart.py
@@ -77,7 +77,7 @@ class Gateway(asyncio.Protocol):
             LOGGER.error("UNKNOWN FRAME RECEIVED: %r", data)  # TODO
 
     def data_frame_received(self, data):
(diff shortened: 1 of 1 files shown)


rcloran:
It would be preferable to open a separate PR for the extra comments.

rcloran:
I'm curious why you changed from .hex() to binascii.hexlify().

rcloran:
Apparently by resolving conflicts here on the PR, I've updated your branch, which was surprising to me.

AndreasBomholtz:
It was because the tests failed for the old python version. So I changed it
to something that worked for all three versions.

Den 3. maj 2017 17.25 skrev "Russell Cloran" <notifications@github.com>:

rcloran:
Ah, OK, that makes sense. I didn't check the previous travis results.

Thanks!

PR #7, opened 2017-05-04
Make sure that the RSTACK frame is the correct one

AndreasBomholtz:
I have added a check to the RSTACK handler, so that it check if the frame is the reply to our reset request.

The reason for this is that when my board starts up, it sends a RSTACK frame with the code "RESET_POWER_ON" and then the program hangs.


bellows/uart.py
```diff
@@ -105,13 +107,15 @@ def rst_frame_received(self, data):
 
     def rstack_frame_received(self, data):
         """Reset acknowledgement frame receive handler"""
-        LOGGER.debug("RSTACK frame: %s", binascii.hexlify(data))
+        LOGGER.debug("RSTACK Version: %d Reason: %s frame: %s", data[1], t.NcpResetCode(data[2]).name, binascii.hexlify(data))
```
rcloran: Since you're creating an `NcpResetCode` twice, I think it might be better to put that in a variable.

AndreasBomholtz: Done

tests/test_uart.py
```diff
@@ -93,9 +93,16 @@ def test_rst_frame_received(gw):
 
 
 def test_rstack_frame_received(gw):
-    gw._reset_future = mock.MagicMock()
```
rcloran: Why are these changes necessary?

Otherwise, this looks good to go.

bellows/uart.py
```diff
@@ -105,13 +107,16 @@ def rst_frame_received(self, data):
...
     def rstack_frame_received(self, data):
         """Reset acknowledgement frame receive handler"""
-        LOGGER.debug("RSTACK frame: %s", binascii.hexlify(data))
         self._send_seq = 0
         self._rec_seq = 0
-        if self._reset_future is None:
-            LOGGER.warn("Reset future is None")
-            return
-        self._reset_future.set_result(True)
+        code = t.NcpResetCode(data[2])
```
rcloran: Hrm. I've just looked at the docs again, and I see that it says "> 0x80: EM3XX-specific error reset code". I wonder if there's a way that we can make this code safe for other reset codes, rather than throwing an exception.

I haven't thought about this too hard, but I think

```python3
try:
    code = t.NcpResetCode(data[2])
except:
    code = t.NcpResetCode.RESET_UNKNOWN_REASON
```

What do you think?

AndreasBomholtz: Yes it makes sense. I have added a new enum for the EM3XX codes.

tests/test_uart.py
AndreasBomholtz: It was a left over from, when I tried to clear the reset_future after the result was set.

PR #8, opened 2017-05-04
Updated client with new commands and parameter handling

AndreasBomholtz:
Added new console commands (command, commands, write_attribute)
Parameters are now checked to see if they are valid for the device.

I hope you like these changes. But just let me know if I should change something.


rcloran:
Thanks!

PR #9, opened 2017-05-05
Reset send_req and rec_seq when receiving a RST ACK.

AndreasBomholtz:
After a reset the counters needs to be reset, or else the sequence is out of order.


rcloran:
Thanks! This probably fixes the occasional problem I've been having where bellows appears to "hang".

PR #10, opened 2017-05-05
Now all endpoint information is loaded from the device if the first command fails.

AndreasBomholtz:
Re-init the device and endpoints if the command to read the information failed.


bellows/zigbee/endpoint.py
```diff
@@ -30,51 +30,59 @@ def __init__(self, device, endpoint_id):
...
-        except:
-            pass
-
-        for cluster in sd.input_clusters:
-            self.add_cluster(cluster)
-
-        self.output_clusters = sd.output_clusters
-
-        self.status = Status.ZDO_INIT
+        if self.status == Status.NEW:
```
rcloran: I'd prefer this to be

```python3
if self.status != Status.NEW:
    return

... do all the other stuff ...
```

bellows/zigbee/endpoint.py
```diff
@@ -59,22 +61,29 @@ def initialize(self):
 
         self.output_clusters = sd.output_clusters
 
+        for cluster in sd.output_clusters:
+            self.add_cluster(cluster)
+
```
rcloran: I have a branch which provides both input and output clusters in separate attributes. Combining them like this doesn't seem like the right thing to do.

bellows/zigbee/endpoint.py
```diff
@@ -59,22 +61,29 @@ def initialize(self):
...
-        self.clusters[cluster_id] = cluster
-        if hasattr(cluster, 'ep_attribute'):
-            self._cluster_attr[cluster.ep_attribute] = cluster
-        listener = bellows.zigbee.appdb.ClusterPersistingListener(
-            self._device.application._dblistener,
-            cluster,
-        )
-        cluster.add_listener(listener)
+        if cluster_id in self.clusters:
+            cluster = self.clusters[cluster_id]
```
rcloran: Rather 

```python
if cluster_id in self.clusters:
    return self.clusters[cluster_id]

cluster = ...
...
```

bellows/zigbee/endpoint.py
AndreasBomholtz: Yes you are right. I just forgot to delete this before I committed. It was a test to handle the incoming ota frames. 
I will remove it.

rcloran:
Not sure why you conditionally added the listener. I've undone that change in 354d553. If the tests pass, I'm going to land as is in the interest of Getting This Done (sorry for letting it linger for so long) -- please feel free to send another PR if there was a reason for that change.

PR #11, opened 2017-05-07
Fix the read_attribute command

mattsch:


bellows/cli/application.py
```diff
@@ -180,15 +180,14 @@ def read_attribute(ctx, attribute):
...
+    if not v:
         click.echo("Received empty response")
-    for record in v[0]:
-        if record.status == 0:
-            click.echo("%s=%s" % (record.attrid, record.value.value))
-        else:
-            click.echo("Attribute %s not successful. Status=%s" % (
-                record.attrid, record.status
-            ))
+    if attribute not in v[0]:
```
AndreasBomholtz: Should this not be a elif?

mattsch: Indeed it should. Fixed now.

rcloran:
Thanks!

PR #13, opened 2017-05-08
Fixed problem with write_attribute.

AndreasBomholtz:
Why did you have `if ad is Analog` in DATA_TYPE_IDX?


bellows/zigbee/zcl/__init__.py
```diff
@@ -192,13 +192,22 @@ def write_attributes(self, attributes):
         for attrid, value in attributes.items():
             if isinstance(attrid, str):
                 attrid = self._attridx[attrid]
+            if attrid not in self.attributes:
+                self.error("%d is not a valid attribute id" % attrid)
```
rcloran: Formatting to logging should be with extra arguments, not by doing the formatting here -- that way a problem in doing the interpolation won't blow up the actual program, and the logging library will print out the error (without any extra handling on the client end).

bellows/zigbee/zcl/__init__.py
```diff
@@ -238,7 +247,7 @@ def _update_attribute(self, attrid, value):
         self.listener_event('attribute_updated', attrid, value)
 
     def log(self, lvl, msg, *args):
-        msg = '[0x%04x:%s:0x%04x] ' + msg
+        msg = '[0x%04x:%s:0x%04x] ' + str(msg)
```
rcloran: I prefer not to do this. I'll put the comments for it where you log the exception.

bellows/zigbee/zcl/__init__.py
```diff
@@ -192,13 +192,22 @@ def write_attributes(self, attributes):
...
-            a.value.value = python_type(value)
-            args.append(a)
+
+            try:
+                python_type = self.attributes[attrid][1]
+                a.value.type = t.uint8_t(foundation.DATA_TYPE_IDX[python_type])
+                a.value.value = python_type(value)
+                args.append(a)
+            except ValueError as e:
+                self.error(e)
```
rcloran: I'm guessing the `str(msg)` in `log` is to handle this. I think you should either stringify the exception here, or add an exception logging method to the `LocalLogMixin`, that matches the logging module's exception... Pretend that the methods are strongly typed and need a string for `msg`.

I think ...

bellows/zigbee/zcl/foundation.py
```diff
@@ -101,7 +101,6 @@ class Discrete:
 DATA_TYPE_IDX = {
     t: tidx
     for tidx, (tname, t, ad) in DATA_TYPES.items()
-    if ad is Analog
```
rcloran: Hrm  - I don't recall what part of the spec I put this in for. I'll need some time to checkout where DATA_TYPE_IDX is otherwise used.

bellows/zigbee/zcl/__init__.py
AndreasBomholtz: Done

bellows/zigbee/zcl/foundation.py
AndreasBomholtz: Okay

bellows/zigbee/zcl/__init__.py
AndreasBomholtz: Done

bellows/zigbee/zcl/__init__.py
AndreasBomholtz: Done

bellows/zigbee/zcl/foundation.py
rcloran: The bit of the spec I was thinking of is:

"2.4.7.1.7 Reportable Change Field
The reportable change field shall contain the minimum change to the attribute that
will result in a report being issued. This field is of variable length. For attributes
with 'analog' data type (see Table 2.16) the field has the same data type as the
attribute. The sign (if any) of the reportable change field is ignored.
For attributes of 'discrete' data type (see Table 2.16) this field is omitted."

It seems that I kinda handled that all properly elsewhere.

I think the problem with this change is that we now run into the problem that line 106 here tries (poorly) to fix -- DATA_TYPES is a many-to-one mapping, so now it's undetermined what, for example, t.uint64_t maps to - is it 0x19, 0x21, 0x31, 0xe8, or 0xe9.

What types in particular were you making this change for? Maybe the answer is to handle them like I've handled t.EmberEUI64?

AndreasBomholtz: I was trying to write a boolean on an On/Off cluster.

Should these discrete values be handled in a other way?

rcloran: I can think of a few possible approaches here. I think your approach is buggy because of the many-to-one mapping.

Other possible approaches:

 - change all the clusters to include the Python type and the ZigBee type. This is probably the only correct way to do it.
 - Keep adding to the special list I've created on line 106+
 - Go over DATA_TYPES twice. In the first pass, put all the analog values in DATA_TYPE_IDX. In the second pass, only put discrete values in if there's no existing entry for that data type in DATA_TYPE_IDX already. This way would probably be my preference for trading off completeness and expedience.

Maybe there is a better way?

AndreasBomholtz: Yes you are right, I have just tested with a t.uint16_t type and I get INVALID_DATA_TYPE back.

So I also think that the only correct way is option 1. 
Because option 2 will not work, because we will still hit the many-to-one issue.
And options 3 will have the wrong data type, if it is not the first one in the list.

So how should we solve this?

rcloran: I'm OK with 2 or 3 as part of this PR. 2 would be deterministic. I'm worried that different implementations of dict in different versions of Python would cause 3 to work differently, unless you sorted by key or something in the comprehension. 2 would be more verbose, though. Up to you.

If you decide to do 2 or 3, please create an issue to remind me that 1 needs to be done at some point :)

PR #15, opened 2017-05-23
Corrected format string

oafilipoai:
This is generating an exception which prevents the home-assistant frontend from starting

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -218,7 +218,7 @@ class ControllerApplication(bellows.zigbee.util.ListenableMixin):
     def _handle_frame_failure(self, message_type, destination, aps_frame, message_tag, status, message):
         try:
             send_fut, reply_fut = self._pending.pop(message_tag)
-            send_fut.set_exception(Exception("Message send failure: %s" % (status, )))
+            send_fut.set_exception(Exception("Message send failure: %d" % (status, )))
             reply_fut.cancel()
         except KeyError:
             LOGGER.warning("Unexpected message send failure")


rcloran:
Could you give me a few more details, please? Specifically, I'd like to know what the actual exception you're seeing is, and what version of Python you're using.

This change makes the logging a little harder to read by a human, by turning it from a named value into an integer. I also don't understand why %s won't work on any value -- it's equivalent to str(), and if that's broken on EzspStatus objects, I'd rather fix that.

oafilipoai:
I'm using Python 3.4.2 (default in debian jessie)

This was the actual exception:
2017-05-18 10:02:52 ERROR (MainThread) [bellows.ezsp] Exception running handler
  File "/home/homeassistant/.homeassistant/deps/bellows/ezsp.py", line 175, in handle_callback
    handler(*args)
  File "/home/homeassistant/.homeassistant/deps/bellows/zigbee/application.py", line 149, in ezsp_callback_handler
    self._handle_frame_failure(*args)
  File "/home/homeassistant/.homeassistant/deps/bellows/zigbee/application.py", line 221, in _handle_frame_failure
    send_fut.set_exception(Exception("Message send failure: %s" % (status, )))
  File "/usr/lib/python3.4/enum.py", line 464, in __str__
    return "%s.%s" % (self.__class__.__name__, self._name_)
AttributeError: 'EmberStatus' object has no attribute '_name_'

rcloran:
Thanks - I think this is fixed by https://github.com/rcloran/bellows/commit/72bc167a1b8752973b20ab04c4f9f2dc58acf278

oafilipoai:
You are right. After I made the changes in that commit the issue went away

rcloran:
Thanks for the report. I'll put up a new version soon.

Issue #18, opened 2017-06-09
Invalid EmberOutgoingMessageType on permit (EZSP 5)


rcloran:
https://community.home-assistant.io/t/one-zigbee-gateway-to-rule-them-all/16803/30

Exception:

```
ERROR:homeassistant.core:Error doing job: Exception in callback SerialTransport.readready()
Traceback (most recent call last):
File "/usr/lib/python3.4/asyncio/events.py", line 120, in _run
self._callback(*self.args)
File "/opt/hass/.homeassistant/deps/serial_asyncio/__init__.py", line 106, in readready
self.protocol.data_received(data)
File "/opt/hass/.homeassistant/deps/bellows/uart.py", line 61, in data_received
self.frame_received(frame)
File "/opt/hass/.homeassistant/deps/bellows/uart.py", line 73, in frame_received
self.data_frame_received(data)
File "/opt/hass/.homeassistant/deps/bellows/uart.py", line 94, in data_frame_received
self._application.frame_received(self._randomize(data[1:-3]))
File "/opt/hass/.homeassistant/deps/bellows/ezsp.py", line 179, in frame_received
result, data = t.deserialize(data, schema)
File "/opt/hass/.homeassistant/deps/bellows/types/__init__.py", line 9, in deserialize
value, data = type_.deserialize(data)
File "/opt/hass/.homeassistant/deps/bellows/types/basic.py", line 13, in deserialize
r = cls(int.from_bytes(data[:cls._size], 'little', signed=cls._signed))
File "/usr/lib/python3.4/enum.py", line 222, in __call__
return cls.__new__(cls, value)
File "/usr/lib/python3.4/enum.py", line 457, in __new__
raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 6 is not a valid EmberOutgoingMessageType
```

MindrustUK:
Hi,

As requested output below:

```
debug: Using selector: EpollSelector
debug: Loading application state from zigbee.db
debug: Sending: b'1ac038bc7e'
debug: RSTACK Version: 2 Reason: RESET_SOFTWARE frame: b'c1020b0a527e'
debug: Send command version
debug: Sending: b'004221a850ed2c7e'
debug: Data frame: b'0142a1a8512805ebd87b7e'
debug: Sending: b'8160597e'
debug: Application frame 0 (version) received
debug: Switching to eszp version 5
debug: Send command version
debug: Sending: b'7d31432157542a1025c57e'
[...]

PR #24, opened 2017-06-28
Added new Ember Outgoing Message Type.

AndreasBomholtz:
I found these new Ember Outgoing Message Type values in the Silabs Gateway Software. This is the missing types for issue #18.


rcloran:
This is slightly concerning to me... it's not clear to me what versions of what firmware use what constants to mean what, or if/when/why constants were re-numbered.

That said, it'll unblock stuff by stopping exceptions from being thrown so... yolo.

Thanks for the PR :)

PR #34, opened 2017-07-24
Make "leave" not remove device; add "remove"

rcloran:
It's normal for a device to leave the network and rejoin. "Leave" shouldn't
remove a device from the application (or the database).


rcloran:
@AndreasBomholtz - thoughts?

AndreasBomholtz:
I think that it looks good.

jkl1337:
Curious as to the status of this. Right now I'm running using the master branch with this merged in with HASS, since otherwise I get disappearing devices on a not terribly infrequent basis. Is this not ready for primetime?

PR #38, opened 2017-08-10
Permit join with key

AndreasBomholtz:
I have implement the support for include with install key from ZigBee 3.


bellows/zigbee/util.py
```diff
@@ -114,3 +116,81 @@ def wrapper(*args, tries=tries, delay=delay, **kwargs):
...
+    (result_len, result) = aesMmoHashUpdate(result_len, result, temp)
+
+    key = t.EmberKeyData()
+    key.contents = t.fixed_list(16, t.uint8_t)(
+        [t.uint8_t(c) for c in result]
+    )
+    return key
+
+
+def convertInstallCode(code):
```
rcloran: PEP-8 compliant name, please.

bellows/zigbee/util.py
```diff
@@ -114,3 +116,81 @@ def wrapper(*args, tries=tries, delay=delay, **kwargs):
...
+        for i in range(AES.block_size):
+            result[i] ^= bytes(data[:AES.block_size])[i]
+
+        data = data[AES.block_size:]
+        length += AES.block_size
+
+    return (length, result)
+
+
+def aesMmoHash(data):
```
rcloran: PEP-8 compliant name, please.

bellows/zigbee/util.py
```diff
@@ -114,3 +116,81 @@ def wrapper(*args, tries=tries, delay=delay, **kwargs):
 
 
 retryable_request = retryable((DeliveryError, asyncio.TimeoutError))
+
+
+def aesMmoHashUpdate(length, result, data):
```
rcloran: PEP-8 compliant name, please.

bellows/zigbee/util.py
```diff
@@ -114,3 +116,81 @@ def wrapper(*args, tries=tries, delay=delay, **kwargs):
...
+
+        data = data[AES.block_size:]
+        length += AES.block_size
+
+    return (length, result)
+
+
+def aesMmoHash(data):
+    result_len = 0
+    remainingLength = 0
```
rcloran: A few variables in here need PEP-8 compliant names, too.

bellows/types/named.py
```diff
@@ -1405,6 +1406,14 @@ class EmberKeyStatus(basic.uint8_t, enum.Enum):
     TC_FAILED_TO_SEND_APP_KEYS = 0x0F
     TC_FAILED_TO_STORE_APP_KEY_REQUEST = 0x10
     TC_REJECTED_APP_KEY_REQUEST = 0x11
+    TC_FAILED_TO_GENERATE_NEW_KEY = 0x12
```
rcloran: Are these from a new version of EZSP? Any other changes from that version that should go into bellows?

bellows/ezsp.py
```diff
@@ -143,7 +143,7 @@ def cb(frame_name, response):
 
     def __getattr__(self, name):
         if name not in self.COMMANDS:
-            raise AttributeError
+            raise AttributeError("%s not found in COMMANDS" % name)
```
rcloran: `name` is unknown input here. It could be a tuple, which could lead to:

```
>>> "This could blow up %s" % ("unexpectedly", "hah")
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
```

It'd be safer to write this as

  raise AttributeError("%s not found in COMMANDS" % (name, ))

bellows/zigbee/util.py
AndreasBomholtz: Done

bellows/zigbee/util.py
AndreasBomholtz: Done

bellows/zigbee/util.py
AndreasBomholtz: Done

bellows/zigbee/util.py
AndreasBomholtz: Done

bellows/ezsp.py
AndreasBomholtz: Yes you are right, fixed.

bellows/types/named.py
AndreasBomholtz: I do not know what version they were added in, but I just looked through "SimplicityStudio_v4/developer/sdks/gecko_sdk_suite/v1.1/protocol/zigbee_5.10/stack/include/ember-types.h" and added what I found.

tests/test_application.py
```diff
@@ -2,7 +2,6 @@
 from unittest import mock
 
 import pytest
-
```
rcloran: Can you add this back, please 

```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

 - http://legacy.python.org/dev/peps/pep-0008/#imports

bellows/types/named.py
rcloran: Have you actually seen these responses from the NCP?

tox.ini
```diff
@@ -12,6 +12,9 @@ setenv = PYTHONPATH = {toxinidir}
 install_command = pip install --process-dependency-links {opts} {packages}
 commands = py.test --cov --cov-report=
 deps =
+    crccheck
```
rcloran: Why do these need to be added here? My understanding is that this list is just for extra test-time dependencies -- tox should be running a pip install of bellows after installing these. I'd prefer to keep the information about what bellows itself depends on in only one place.

tests/test_application.py
AndreasBomholtz: Done

tox.ini
AndreasBomholtz: Yes, you are right. I have removed them again. I think I added them while I was testing.

bellows/zigbee/application.py
```diff
@@ -49,6 +49,8 @@ def initialize(self):
         yield from self._cfg(c.CONFIG_APPLICATION_ZDO_FLAGS, zdo)
         yield from self._cfg(c.CONFIG_TRUST_CENTER_ADDRESS_CACHE_SIZE, 2)
         yield from self._cfg(c.CONFIG_PACKET_BUFFER_COUNT, 0xff)
+        yield from self._cfg(c.CONFIG_KEY_TABLE_SIZE, 1)
+        yield from self._cfg(c.CONFIG_TRANSIENT_KEY_TIMEOUT_S, 180)
```
jimmo: My HUSBZB-1 doesn't seem to support CONFIG_TRANSIENT_KEY_TIMEOUT_S   (EzspStatus.ERROR_INVALID_ID)

The device reports as EZSP version 4. I'm not aware of any way to do a firmware update?

PR #41, opened 2017-08-11
The baudrate for the uart can now be changed

AndreasBomholtz:
Added optional baudrate for uart.


bellows/uart.py
```diff
@@ -249,9 +249,11 @@ def _unstuff(self, s):
 
 
 @asyncio.coroutine
-def connect(port, application, loop=None):
+def connect(port, application, loop=None, baudrate=None):
     if loop is None:
         loop = asyncio.get_event_loop()
+    if baudrate is None:
```
rcloran: Instead of making the kwarg default to None and doing this, just make it default to 57600.

AndreasBomholtz: But I'm passing in "None" from the cli and I'm always assigning the baudrate, so we will never use the default value, because it is always overwritten.
Should I try to rework the other layers so that they check if the baudrate is None, before calling the uart connect?

rcloran: Right.

bellows/cli/util.py
```diff
@@ -86,11 +87,11 @@ def channel_mask(channels):
 
 
 @asyncio.coroutine
-def setup(dev, cbh=None, configure=True):
+def setup(dev, cbh=None, configure=True, baudrate=None):
```
rcloran: Let's make baudrate required here (and put it together with dev, but it'll have to be once you kill the default value). Also remove it being kwarg-like in all the calls to it.

bellows/cli/util.py
```diff
@@ -112,9 +113,9 @@ def cfg(config_id, value):
 
 
 @asyncio.coroutine
-def setup_application(dev, database_file, startup=True):
+def setup_application(dev, database_file, startup=True, baudrate=None):
     s = bellows.ezsp.EZSP()
-    yield from s.connect(dev)
+    yield from s.connect(dev, baudrate=baudrate)
```
rcloran: Can you change either this call or the call in setup() above so that they use the same style.

bellows/cli/util.py
```diff
@@ -112,9 +113,9 @@ def cfg(config_id, value):
 
 
 @asyncio.coroutine
-def setup_application(dev, database_file, startup=True):
+def setup_application(dev, database_file, startup=True, baudrate=None):
```
rcloran: Make it required here, too (and next to dev).

AndreasBomholtz: I will make it required.

bellows/cli/util.py
AndreasBomholtz: Should there be a default value in the cli, so that the use do not have to supply it, but they can change it.

PR #60, opened 2017-10-08
Add error reporting when database isn't provided when needed

halkeye:
New output:
```
Usage: bellows devices [OPTIONS]

Error: Invalid value for "-D" / "--database": Path "/home/halkeye/.config/bellows/app.db" does not exist.
```

### DIFF
--- a/bellows/cli/opts.py
+++ b/bellows/cli/opts.py
@@ -38,10 +38,13 @@ baudrate = click.option(
 database_file = click.option(
     '-D', '--database',
     type=click.Path(
+        exists=True,
         dir_okay=False,
         writable=True,
     ),
+    required=True,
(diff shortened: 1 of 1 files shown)


halkeye:
Looks like the CI broke on installing six for linting

I just tried it locally (`tox -e lint`) and it had no issues.

rcloran:
Without this commit, the database is automatically initialized when it doesn't exist -- for example, when setting up a new network/adding the first node to a new network. That makes it easier to get going... but, I've received enough "bugs" related to this that maybe checking its existence is the right way to go. Let's try it.

halkeye:
If the directory doesn't exist it spits out a large stack trace with no indication on what the path should be.

rcloran:
Thanks!

PR #61, opened 2017-10-11
Added custom click type to handle node parameter

AndreasBomholtz:


bellows/cli/util.py
```diff
@@ -24,6 +24,15 @@ def convert(self, value, param, ctx):
         return values
 
 
+class ZigbeeNodeParamType(click.ParamType):
+    name = 'collon separated hex bytes'
```
rcloran: Spelling: “colon”

Issue #58, opened 2017-09-27
ValueError: 153 is not a valid EzspStatus


halkeye:
Not sure what is useful for debugging, so including the entire stack trace.

I'm really just playing around with the HUSBZB-1 and seeing if i can't add some random zigbee devices I have.

```
(homeassistant) homeassistant@hassbian:~/bellows $ bellows -v debug -d /dev/zigbee permit -D ~/.homeassistant/zigb
ee.db
debug: Using selector: EpollSelector
debug: Loading application state from /home/homeassistant/.homeassistant/zigbee.db
debug: Sending: b'1ac038bc7e'
debug: RSTACK Version: 2 Reason: RESET_SOFTWARE frame: b'c1020b0a527e'
debug: Send command version
debug: Sending: b'004221a850ed2c7e'
debug: Data frame: b'0142a1a8502805e67f627e'
debug: Sending: b'8160597e'
debug: Application frame 0 (version) received
debug: Send command setConfigurationValue
debug: Sending: b'7d314321fb582815c3277e'
debug: Data frame: b'1243a1fb54fb737e'
debug: Sending: b'82503a7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
debug: Sending: b'224021fb592f15226f7e'
debug: Data frame: b'2340a1fb54c6107e'
debug: Sending: b'83401b7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
debug: Sending: b'334121fb792b15a2d77e'
debug: Data frame: b'3441a1fb54d32a7e'
debug: Sending: b'8430fc7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
debug: Sending: b'444621fb7d5e291514417e'
debug: Data frame: b'4546a1fb5435d07e'
debug: Sending: b'8520dd7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
debug: Sending: b'554721fb4d2815713f7e'
debug: Data frame: b'5647a1fb54a9ec7e'
debug: Sending: b'8610be7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
debug: Sending: b'664421fb55d515b18f7e'
debug: Data frame: b'6744a1fb54948f7e'
debug: Sending: b'87009f7e'
debug: Application frame 83 (setConfigurationValue) received
debug: Send command setConfigurationValue
[...]

PR #64, opened 2017-10-16
Check crc in received frames

AndreasBomholtz:
Verify received CRC.

I'm sending a NAK frame back to the NCP, but I do not know it this will solve the issue #58. I have just tried with a normal frame and the NCP resends the frame. So what I'm afraid off, is that it just keeps sends the same frame with CRc error and we then gets stuck.

### DIFF
--- a/bellows/uart.py
+++ b/bellows/uart.py
@@ -64,7 +64,15 @@ class Gateway(asyncio.Protocol):
         """Extract a frame from the data buffer"""
         if self.FLAG in data:
             place = data.find(self.FLAG)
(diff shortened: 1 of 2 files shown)


rcloran:
I'm uneasy about ACK after 3 invalid frames. I'd like to do the check and log+nak -- we'll hear about it if it creates any more issues, and have more debugging info, at least. #58 seems like it was caused by something else, in the end, though?

If you don't have time to, you OK with me making the change to remove the 3 NAK counter?

AndreasBomholtz:
I have removed the check now.

Yes lets just way and see if anybody will hit this issue.

PR #79, opened 2017-12-11
Add support for Xiaomi, Trådfri

jonatanolofsson:
I have tried to separate this PR in several commits, however some changes rely on previous commits so I submit it as a single PR for review.

In short I started learning the code a few days ago, and started cleaning as I went - starting to get rid of magic numbers etc, which I am allergic to  =)

I also wanted a way to not use Home Assistant, so I started a base for a standalone mode. The original mode was to mimick the deconz-rest-plugin for ConBee with a websocket for updates and a (Hue compatible) REST interface for control. Not sure where that is going, but it's most definetly useful for debugging, as I can startup bellows as a self-contained server which I can interact with with curl-commands.

In order to deal with Xiaomi sensors (door sensors is what I tested with), I alleviated the demand that all endpoints and clusters must be set up by the time of joining - Xiaomi doesn't report those, but are sent anyway. This is handled by making the message handling asynchronous, and adding missing endpoints/clusters as they present themselves - instead of just warning about it. Xiaomi also uses the Structure zigbee data field, so I added a parser for arrays, structures, set and bags.

In order to support Trådfri devices, I made a minor specialization in the code for IKEA devices, setting profile_id in the aps frame to ZHA even though it's ZLL. For this, I needed to store the manufacturer in the database and since I was already at it I did some cleanup of the database schema - adding device manufacturer and removing device and endpoint status (non-initialized devices/endpoints should not be saved). 

Instead of adding a framework for specialization, I opted to keep it simple and add just the code needed. I think future specializations will reveal how a more generic handling could look, but considering all possible variations is futile at this point.

### DIFF
--- a/bellows/cli/application.py
+++ b/bellows/cli/application.py
@@ -32,6 +32,25 @@ def form(ctx, database, channel, pan_id, extended_pan_id):
 
     return util.app(inner, app_startup=False)(ctx)
 
+@main.command()
+@opts.database_file
+@opts.wshost
+@opts.wsport
+@opts.resthost
+@opts.restport
+@opts.restapikey
+@click.pass_context
+def standalone(ctx, database, wshost, wsport, resthost, restport, rest_api_key):
+    """Run a standalone server."""
+    import bellows.standalone
+    ctx.obj['database_file'] = database
+    ctx.obj['wshost'] = wshost
+    ctx.obj['wsport'] = wsport
+    ctx.obj['resthost'] = resthost
+    ctx.obj['restport'] = restport
+    ctx.obj['rest_api_key'] = rest_api_key
+
+    return util.app(bellows.standalone.start, app_startup=False, run_forever=True)(ctx)
 
 @main.command()
 @opts.database_file
--- a/bellows/cli/application.py
+++ b/bellows/cli/application.py
@@ -8,6 +8,8 @@ import click
 import bellows.ezsp
 import bellows.zigbee.application
 import bellows.zigbee.endpoint
+import bellows.types as t
+from bellows.zigbee.zdo.types import CLUSTER_ID as ZDO_CLUSTER_ID
 from . import opts
 from . import util
 from .main import main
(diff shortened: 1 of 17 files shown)


dmulcahey:
@jonatanolofsson 

```
     def _save_endpoints(self, device):
         self.execute("DELETE FROM endpoints WHERE ieee = ?", (device.ieee, ))
  		  
     def _save_input_clusters(self, endpoint):
        self.execute("DELETE FROM input_clusters WHERE ieee = ?", (endpoint.device.ieee, ))
```

These 2 delete lines seem to be preventing home assistant from starting a 2nd time after these changes are applied. Are they necessary? The subsequent queries does an insert or replace.

jonatanolofsson:
I'm not sure about your exact errors, but I just fixed some db issues on my [develop-branch](https://github.com/jonatanolofsson/bellows/tree/develop) that may have caused this. That one is rebased on @AndreasBomholtz develop-branch, which together makes for quite a few commits past the master branch of @rcloran . 

The point of the delete was to do a proper refresh and make sure the database was consistent with the endpoint while I was tinkering - especially the Xiaomi devices are notorious for not reporting all clusters, so I've been experimenting with that. In a more stable environment, it might make more sense to keep all clusters ever created, so you should be able to remove those deletes.

I'm going to close this PR since I do not intend to maintain it, but should @rcloran want parts of it, there's some mergework to do. My development branch has moved too far away by now to be backwards compatible.

dmulcahey:
@jonatanolofsson I have actually been watching and tinkering with the changes that you guys have been making (@AndreasBomholtz and yourself). Do you have any interest in forming a Discord group / channel to facilitate conversation around this stuff? I have some questions and some ideas.

jonatanolofsson:
I tried to propse that myself to @rcloran some time ago actually. I just created a group @ [https://discord.gg/D5dRgXQ](https://discord.gg/D5dRgXQ)

PR #65, opened 2017-10-17
Handle new id in join

AndreasBomholtz:
If a device changes ID during join we will try to talk to the new ID instead. 
I'm trying to handle the issue #44

PR #83, opened 2017-12-22
Persisting new network addressed

ilmaruk:
I noticed that if a device rejoins with a new network address (say it's been manually factory reset), its new network address doesn't get persisted (it's only updated in memory), so that anything coming from the device after the app is restarted, is deemed as "from unknown device".
This PR address this limitation, but it's my very first contribution, so please let me know if I should have approached this in a different way; thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ ENV/
 
 # Editor temp files
 .*.swp
+
+# IDEs stuff
+.idea/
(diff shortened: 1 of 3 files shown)


dmulcahey:
Just pointing out: #65 looks similar

rcloran:
If there are any further changes you suggest around this code, please open up the PR on zigpy (https://github.com/zigpy/zigpy).

PR #89, opened 2018-01-21
bug-fix: use the remote source endpoint id to direct msgs to right endpoints

Yoda-x:
this patch uses the source endpoint in a incoming packet to assign the packet to right endpoint.
example: getting attribute reports from device , endpoint 10, cluster 0x702, send to endpoint 1
current code -> incoming packet from source-endpoint 10, cluster x, destination-endpoint 1 -> unknown cluster 0x702 for endpoint 1.
with this patch the message goes to the correct endpoint 10.
this shows only when the device uses an endpoint != 1, why it may have not come to attention before

### DIFF
--- a/bellows/zigbee/device.py
+++ b/bellows/zigbee/device.py
@@ -97,11 +97,11 @@ class Device(zutil.LocalLogMixin):
 
     def handle_message(self, is_reply, aps_frame, tsn, command_id, args):
         try:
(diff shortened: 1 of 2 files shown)


rcloran:
I'm happy to accept a change that looks like this. The code has been moved to zigpy (https://github.com/zigpy/zigpy). Please do open a PR there.

PR #71, opened 2017-11-09
Add missing stuff for thermostat support

titilambert:

PR #91, opened 2018-01-29
Fix fans to use enums for attributes.

igorbernstein2:
I'm trying to get bellows to communicate with 
https://www.homedepot.com/p/Hampton-Bay-Universal-Wink-Enabled-White-Ceiling-Fan-Premier-Remote-Control-99432/206591100.

More details about the receiver can be found here:
https://community.smartthings.com/t/home-decorators-ceiling-fan-light-controller-mr101z-first-impressions/81408

And the smartthings implementation can be found here:
https://github.com/dcoffing/KOF-CeilingFan/blob/master/devicetypes/dcoffing/kof-zigbee-fan-controller.src/kof-zigbee-fan-controller.groovy

There are few issues to work through, this solves the first one, writing the ceiling fan attribute results in the following error:
```[[<WriteAttributesStatusRecord status=Status.INVALID_DATA_TYPE attrid=0>]]```

As described in the ZCL [cluster spec](http://www.zigbee.org/~zigbeeor/wp-content/uploads/2014/10/07-5123-06-zigbee-cluster-library-specification.pdf) `6.4.2.2 Attributes`, the 2 available attributes are both enums. I'm following the approach taken in #71.

### DIFF
--- a/bellows/types/basic.py
+++ b/bellows/types/basic.py
@@ -85,6 +85,9 @@ class uint64_t(uint_t):  # noqa: N801
 class enum8(uint8_t):  # noqa: N801 
     pass
 
+class enum16(uint16_t):  # noqa: N801 
+    pass
+
 class Single(float):
     def serialize(self):
         return struct.pack('<f', self)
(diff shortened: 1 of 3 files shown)


rcloran:
Please open this PR again in zigpy (https://github.com/zigpy/zigpy). Unfortunately it seems there's no way to move stuff across.

PR #93, opened 2018-02-06
Added manufacturer support in cli

AndreasBomholtz:
This will add manufacturer support when zigpy has that feature.


AndreasBomholtz:
Now that zigpy supports manufacturer commands, this can be merged now.

PR #100, opened 2018-02-12
Using names types instead of numbers

### LINKED ISSUE #98: Remove magic numbers from source code
Currently, there are a lot of magic numbers in the source which could be removed. Use names/constants instead of numbers when e.g. referring to a command.

AndreasBomholtz:
This is my attempt to fix #98. What do you say @jonatanolofsson ?

### DIFF
--- a/bellows/cli/application.py
+++ b/bellows/cli/application.py
@@ -5,6 +5,7 @@ import binascii
 
 import click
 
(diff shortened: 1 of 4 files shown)


jonatanolofsson:
Nice Andreas. Looks good.

Issue #76, opened 2017-11-16
Error trying to pair thermostat devices (st218)


titilambert:
Hello,

I got devices paired, but I wanted to retry the pairing.
So I stopped my Home assistant, delete zigbee.db file et restart Home assistant.
Now When I try to pair devices I got that
```
ov 16 13:16:51 androlapin hass[24814]: 2017-11-16 13:16:51 WARNING (MainThread) [bellows.zigbee.zdo] [0xe895:zdo] Unsupported ZDO request 0x0036
Nov 16 13:16:51 androlapin hass[24814]: 2017-11-16 13:16:51 WARNING (MainThread) [bellows.zigbee.zdo] [0xe895:zdo] Unsupported ZDO request 0x0036
Nov 16 13:16:51 androlapin hass[24814]: 2017-11-16 13:16:51 WARNING (MainThread) [bellows.zigbee.zdo] [0xe895:zdo] Unsupported ZDO request 0x0036
Nov 16 13:16:51 androlapin hass[24814]: 2017-11-16 13:16:51 ERROR (MainThread) [homeassistant.core] Timer got out of sync. Resetting
Nov 16 13:17:28 androlapin hass[24814]: 2017-11-16 13:17:28 ERROR (MainThread) [homeassistant.core] Timer got out of sync. Resetting
Nov 16 13:18:28 androlapin hass[24814]: 2017-11-16 13:18:28 ERROR (MainThread) [homeassistant.core] Error doing job: Exception in callback SerialTransport._read_ready()
Nov 16 13:18:28 androlapin hass[24814]: Traceback (most recent call last):
Nov 16 13:18:28 androlapin hass[24814]:   File "/usr/lib/python3.6/asyncio/events.py", line 127, in _run
Nov 16 13:18:28 androlapin hass[24814]:     self._callback(*self._args)
Nov 16 13:18:28 androlapin hass[24814]:   File "/opt/homeassistant/deps/lib/python3.6/site-packages/serial_asyncio/__init__.py", line 106, in _read_ready
Nov 16 13:18:28 androlapin hass[24814]:     self._protocol.data_received(data)
Nov 16 13:18:28 androlapin hass[24814]:   File "/opt/homeassistant/deps/lib/python3.6/site-packages/bellows/uart.py", line 61, in data_received
Nov 16 13:18:28 androlapin hass[24814]:     self.frame_received(frame)
[...]

PR #101, opened 2018-02-12
Added guard before setting result on reset future

AndreasBomholtz:
This should fix the issue #76. @titilambert Can you test this small change?

### DIFF
--- a/bellows/uart.py
+++ b/bellows/uart.py
@@ -125,7 +125,9 @@ class Gateway(asyncio.Protocol):
             LOGGER.warn("Reset future is None")
             return
 
-        self._reset_future.set_result(True)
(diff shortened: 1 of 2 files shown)


bellows/uart.py
```diff
@@ -125,7 +125,9 @@ def rstack_frame_received(self, data):
             LOGGER.warn("Reset future is None")
             return
 
-        self._reset_future.set_result(True)
+        # Make sure that the reset_future is not done
+        if self._reset_future.done() is not True:
```
rcloran: `if not self._reset_future.done()` ?

AndreasBomholtz: It was just a easy way, not to rewrite the test cases. But I can just rewrite them to make them pass.

PR #104, opened 2018-02-18
increase table size

Yoda-x:
increased max childen to 32
address-table to 16
source_routes to 8

tried to be conservative to keep enough buffers for normal operations


bellows/zigbee/application.py
```diff
@@ -40,6 +40,9 @@ def initialize(self):
         )
         yield from self._cfg(c.CONFIG_APPLICATION_ZDO_FLAGS, zdo)
         yield from self._cfg(c.CONFIG_TRUST_CENTER_ADDRESS_CACHE_SIZE, 2)
+        yield from self._cfg(c.CONFIG_ADDRESS_TABLE_SIZE, 16)
+        yield from self._cfg(c.CONFIG_SOURCE_ROUTE_TABLE_SIZE, 8)
```
AndreasBomholtz: The default value from Silabs is 32 for the "CONFIG_SOURCE_ROUTE_TABLE_SIZE". Maybe we should at least keep it at that value or higher. I have seen in some of the Silabs example code that they have values like 100 and 250 for this value.

Yoda-x: if you look at the UG100 doc from silab, page 5: it says the default value is 0.
i did also  a config_read before writing for my em357 stick:
2018-02-23 21:53:51 DEBUG (MainThread) [bellows.zigbee.application] default CONFIG_ADDRESS_TABLE_SIZE: 8
2018-02-23 21:53:51 DEBUG (MainThread) [bellows.zigbee.application] default c.CONFIG_NEIGHBOR_TABLE_SIZE: 16
2018-02-23 21:53:51 DEBUG (MainThread) [bellows.zigbee.application] default c.CONFIG_SOURCE_ROUTE_TABLE_SIZE: 0
2018-02-23 21:53:51 DEBUG (MainThread) [bellows.zigbee.application] default c.CONFIG_MAX_END_DEVICE_CHILDREN: 6

I`m pretty sure there a different default settings depending on the OEM or chipset(em357:12kB RAM, em3588:64kB RAM)
When I used larger tables, I saw often failures, where I had no buffers left.
From what I read is the MAX.END_DEVICE_CHILDREN table a hard stop, if it's full it is full.
the other tables do a cleanup or time_out old values.

if you want to check your default settings, thats the code I added https://gist.github.com/Yoda-x/8ab346a3ef6d7c08c9e9750ebdfbef98.js

AndreasBomholtz: I have the same default values in the chip, I was talking about the Silabs Zigbee Gateway program that configures the chip. It configures the values different.

I'm working on a MGM111 which has 32Kb ram, so it works fine for me. But if there is a memory issue with the lower end chips, then we should not increase it.

I will just accept your values.

PR #109, opened 2018-03-11
implements zdo Management Permit Join Request broadcast

Yoda-x:
enables child routers to allow endpoint joins. With this endpoints can join outside the radio range of the controller.
Implements also a zdo broadcast function. if this file is the wrong place, please let me know.

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -257,7 +257,6 @@ class ControllerApplication(zigpy.application.ControllerApplication):
 
     def permit(self, time_s=60):
         assert 0 <= time_s <= 254
-        """ send mgmt-permit-join to all router """
         yield from self.send_zdo_broadcast(0x0036, 0x0000, 0x00, [time_s,0])
         return self._ezsp.permitJoining(time_s)
 
@@ -280,13 +279,11 @@ class ControllerApplication(zigpy.application.ControllerApplication):
         if v[0] != t.EmberStatus.SUCCESS:
             raise Exception("Failed to change policy to allow generation of new trust center keys")
         
-        """ send mgmt-permit-join to all router """
         yield from self.send_zdo_broadcast(0x0036, 0x0000, 0x00, [time_s,0])
         return self._ezsp.permitJoining(time_s, True)
 
     @asyncio.coroutine
     def send_zdo_broadcast(self, command, grpid, radius,   args):
-        """ create aps_frame for zdo broadcast"""
         aps_frame = t.EmberApsFrame()
         aps_frame.profileId = t.uint16_t(0x0000) # 0 for zdo
         aps_frame.clusterId =  t.uint16_t(command)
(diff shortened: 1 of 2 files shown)


Yoda-x:
It seems the tests don't like the call of the sendbroadcast. As it's not the expected behavior.

Yoda-x:
The reason for bellows: the permit call is in bellows and it uses ezsp specific calls and aps_frame. I agree, the functionality is useful for all kind of hardware, but I have only knowledge and hardware for the ezsp kind.  I will modify the comments.

Yoda-x:
I removed the comments, please check if it's ok now.

Yoda-x:
I updated the tests now and added a new function test. 
travis and coverage are both green now. Please review again and state if something is missing

bellows/zigbee/application.py
```diff
@@ -277,4 +279,24 @@ def permit_with_key(self, node, code, time_s=60):
         if v[0] != t.EmberStatus.SUCCESS:
             raise Exception("Failed to change policy to allow generation of new trust center keys")
 
+        yield from self.send_zdo_broadcast(0x0036, 0x0000, 0x00, [time_s, 0])
         return self._ezsp.permitJoining(time_s, True)
+
+    @asyncio.coroutine
+    def send_zdo_broadcast(self, command, grpid, radius, args):
+        aps_frame = t.EmberApsFrame()
+        aps_frame.profileId = t.uint16_t(0x0000)       # 0 for zdo
```
rcloran: Is there any reason that logic like this can't be encapsulated in the ZDO layer? I guess the whole device-oriented API doesn't make it natural to do that, but it still seems to be that adding ZDO-specific stuff into the application layer will make it a little awkward to deal with other kinds of broadcasts later.

I'd certainly be happy to have a `send_broadcast` inside the application. Similar to "request", I think it should be agnostic of whether it's ZCL or ZDO (or anything else that can be dreamed up at that layer).

bellows/zigbee/application.py
```diff
@@ -256,6 +257,7 @@ def request(self, nwk, profile, cluster, src_ep, dst_ep, sequence, data, expect_
 
     def permit(self, time_s=60):
         assert 0 <= time_s <= 254
+        yield from self.send_zdo_broadcast(0x0036, 0x0000, 0x00, [time_s, 0])
```
rcloran: This method is now a little weird -- it's a generator which also returns a coroutine. After #114, I think this should be written something like

```
async def permit(self, ...):
    assert ...
    await self.send_zdo_broadcast(...)
    return await self._ezsp.permitJoining(...)
```

bellows/zigbee/application.py
Yoda-x: Ok, I think this makes sense. let me create a more generic send_broadcast (need this anyway for some multicast/group stuff I'm working on) and create a send_broadcast inside the ZDO section.  
I will keep EZSP specific stuff (APS frame, and ezsp_sendmulticast ) inside bellows and the other functions inside zigpy. 
Ok, stay tuned

bellows/zigbee/application.py
Yoda-x: right, will do this

Yoda-x:
I cancel this as I don't have the option to test it anymore, I assume @Adminiuga will do a pull that fits the current state

PR #125, opened 2018-07-28
WIP: Update to EZSP 6.3

cavokz:
Initial scan of the commands in EZSP 6.3 (from UG100: EZSP Reference Guide)

RF4CE seems gone from the new version, do bellows make any use of those commands?
How different versions of the protocol are going to be handled?
Which version is the current reference and which part of it is actually covered?

### DIFF
--- a/bellows/commands.py
+++ b/bellows/commands.py
@@ -18,6 +18,7 @@ COMMANDS = {
     'setGpioRadioPowerMask': (0xAE, (t.uint32_t, ), ()),
     'setCtune': (0xF5, (t.uint16_t, ), ()),
     'getCtune': (0xF6, (), (t.uint16_t, )),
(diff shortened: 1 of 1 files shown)


duncanfinney:
I had to add another EmberNodeType for no network. 0xFB

Hedda:
@cavokz @duncanfinney Any updates on this? The "UG100: EZSP Reference Guide" has been updated:

https://www.silabs.com/support/resources.ct-manuals.p-wireless_zigbee-and-thread?query=ug100

Version 3.9 of that document was added 2/26/2020 is up to date with EmberZNet PRO Release 6.7.

cavokz:
I'm not sure that this PR has a any sense any more (if it had any at all), maybe it should be reopened as an issue but again, maybe those questions are not so useful in the end. Feel free to close it without merge, I won't complain. Thanks.

PR #135, opened 2018-10-12
Add Zigbee Endpoint to coordinator.

Adminiuga:
Add endpoint containing IAS client cluster during HUSBZB initialization.
A few IAS devices, particularly [Bosch ISW-ZPR1-WP13](https://us.boschsecurity.com/en/products/intrusionalarmsystems/premiseswireless/zigbee/zigbee_products_56458) (humongous, but cheap) and [Nyce-3014-HA](https://www.amazon.com/NYCE-NCZ-3014-HA-ZigBee-Garage-Sensor/dp/B00WTDV1TU) (and Nyce is very nyce) won't stay on the network after joining, unless there's an endpoint with IAS client cluster on the controller. 

Kudos to @Yoda-x for figuring this out. For more details please follow [Support for HA version of NYCE sensors](https://github.com/Yoda-x/ha-zha-new/issues/47) issue on `zha_new`

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -48,6 +48,24 @@ class ControllerApplication(zigpy.application.ControllerApplication):
         await self._cfg(c.CONFIG_END_DEVICE_POLL_TIMEOUT, 60)
         await self._cfg(c.CONFIG_END_DEVICE_POLL_TIMEOUT_SHIFT, 6)
 
+        await self.add_endpoint(
+            output_clusters=[zigpy.zcl.clusters.security.IasZone.cluster_id]
+        )
+
+    async def add_endpoint(self, endpoint=1,
+                           profile_id=zigpy.profiles.zha.PROFILE_ID,
+                           device_id=0xbeef,
+                           app_flags=0x00,
+                           input_clusters=[],
+                           output_clusters=[]):
+        """Add endpoint."""
+        res = await self._ezsp.addEndpoint(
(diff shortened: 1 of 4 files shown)


Adminiuga:
@AndreasBomholtz just a friendly ping. Are there any concerns about merging this?

BTW, same problem does exist with `zigpy-xbee`, but not this `zigpy-deconz`, because by default deconz has an endpoint with 0x0500 cluster ID, so I do believe this is a right fix.

bellows/zigbee/application.py
```diff
@@ -48,6 +48,24 @@ def __init__(self, ezsp, database_file=None):
         await self._cfg(c.CONFIG_END_DEVICE_POLL_TIMEOUT, 60)
         await self._cfg(c.CONFIG_END_DEVICE_POLL_TIMEOUT_SHIFT, 6)
 
+        await self.add_endpoint(
+            output_clusters=[zigpy.zcl.clusters.security.IasZone.cluster_id]
+        )
+
+    async def add_endpoint(self, endpoint=1,
+                           profile_id=zigpy.profiles.zha.PROFILE_ID,
+                           device_id=0xbeef,
```
AndreasBomholtz: Why is device_id _beef_ ?

Adminiuga: it is a `device_type` of the endpoint. Since we're not strictly a `DeviceType.IAS_ZONE` device I've choose some "random" value. But we can change it to something else.

PR #136, opened 2018-11-02
Broadcast unicast join

Adminiuga:
This is an accompanying PR for https://github.com/zigpy/zigpy/pull/91
Implement `zdo.broadcast()` and `ControllerApplication.broadcast()` methods
Rename `ControllerApplication.permit()` to `ControllerApplication.permit_ncp()`

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -4,9 +4,11 @@ import logging
 import os
 
 from zigpy.exceptions import DeliveryError
+from zigpy.types import BroadcastAddress
 import zigpy.application
 import zigpy.device
 import zigpy.util
+import zigpy.zdo
 
 import bellows.types as t
 import bellows.zigbee.util
@@ -257,6 +259,7 @@ class ControllerApplication(zigpy.application.ControllerApplication):
 
         # Wait for messageSentHandler message
         v = await send_fut
+
         if expect_reply:
             # Wait for reply
             try:
@@ -291,3 +294,33 @@ class ControllerApplication(zigpy.application.ControllerApplication):
             raise Exception("Failed to change policy to allow generation of new trust center keys")
 
         return self._ezsp.permitJoining(time_s, True)
+
+    async def broadcast(self, profile, cluster, src_ep, dst_ep, grpid, radius,
+                        sequence, data,
+                        broadcast_address=BroadcastAddress.RX_ON_WHEN_IDLE):
+        assert sequence not in self._pending
+        send_fut = asyncio.Future()
+        self._pending[sequence] = (send_fut, None)
+
+        aps_frame = t.EmberApsFrame()
+        aps_frame.profileId = t.uint16_t(profile)
+        aps_frame.clusterId = t.uint16_t(cluster)
+        aps_frame.sourceEndpoint = t.uint8_t(src_ep)
+        aps_frame.destinationEndpoint = t.uint8_t(dst_ep)
+        aps_frame.options = t.EmberApsOption(
+            t.EmberApsOption.APS_OPTION_NONE
+        )
+        aps_frame.groupId = t.uint16_t(grpid)
+        aps_frame.sequence = t.uint8_t(sequence)
+
+        LOGGER.debug("broadcast: %s - %s", aps_frame, data)
+        v = await self._ezsp.sendBroadcast(broadcast_address, aps_frame,
+                                           radius, sequence, data)
+        if v[0] != t.EmberStatus.SUCCESS:
+            self._pending.pop(sequence)
+            send_fut.cancel()
+            raise DeliveryError("Broadcast failure: %s", v[0])
+
+        # Wait for messageSentHandler message
+        v = await send_fut
+        return v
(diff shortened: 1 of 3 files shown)


bellows/zigbee/application.py
```diff
@@ -249,7 +252,7 @@ def _handle_frame_sent(self, message_type, destination, aps_frame, message_tag,
                 raise
         return v
 
-    def permit(self, time_s=60):
+    def permit_ncp(self, time_s=60):
```
AndreasBomholtz: Why do you need to rename this?
But you also need to change the permit call in _cli/application.py_.

Adminiuga: I can rename permit call in _cli/application.py_, but before doing so, I'd like to take a step back and confirm a few things/assumtions: Lets say a user have formed a Zigbee network consisting of EZSP based radio and two Zigbee router devices. What should the user running `bellows permit` command expect?
1. new zigbee device is allowed to join through any existing router or coordinator
2. new zigbee device is allowed to join only through coordinator

In this PR (in tandem with https://github.com/zigpy/zigpy/pull/91, as we would need to merge both PRs at the same time), running `bellows permit` allows new zigbee device to join the network through any existing router or coordinator.

> Why do you need to rename this?

**Short version:** in current implementation, calling `permit()` opens only the radio for new device joins, so I renamed it to reflect this aspect.

**Long version:** Initially I had a version using the same name, but then we either need to call `super().permit()` for the code  broadcasting ZDO `mgmt_permit_join_req` and call radio specific code opening it to new join requests.

Another option, like in https://github.com/zigpy/zigpy/pull/88 along with `permit_targeted()` we implement the `permit_broadcast()`.  I'm not really a fan of this option because:
- I feel it's way to many different API calls for a single operation
- same criticism as at the top of this message. If I would be just starting with bellows/zigpy, then my natural assumption about `permit()` would be that it opens all devices in the network for join requests
- Other 3rd party apps using zigpy wouldn't benefit from this functionality, unless they update their code as well

That was when @dmulcahey floated the idea of renaming of `permit()` to `permit_ncp()` and after some consideration and POC implementation I've accepted and liked the idea, because:
- method's name actually is reflecting what it does: it opens only radio for new joins
- `permit()` stays at `zigpy` and `permit_ncp()` is implemented by radio library
- we have same, backward compatible interface. We can either do targeted or open entire network for new join requests.
- other 3rd party apps benefit from this PR without any change to their code (backward compatible).
When I started with hass/bellows, one of my very 1st frustrating experience was: I couldn't get new bulbs to join the network even though I had a few other bulbs acting as router in immediate vicinity of a new bulb. I had to move the new bulb next to my radio to join it and I couldn't understand why, because I genuinely believed that running `bellows permit` allows new node to join through any router, because that's how Zigbee should operate from what I was reading at the time :) And having a few 45857GE wall switches installed, carrying radio close to each switch was far from ideal. And if for whatever reason I had to rejoin 45857GE switches, that was a big PITA

This was my train of thoughts behind this PR implementation and I may be missing some other important points. With that said, I leave it up to you as the project maintainer whether we should rename `permit()` or not.

AndreasBomholtz: Thanks for explain the issue. I'm with you and I would expect that permit would use option 1. So your PR makes sense.
I do not think that @rcloran will come back and comment on zigpy/zigpy#91 so I thnik that I will merge them.

AndreasBomholtz:
@Adminiuga Can you push a "change" to your branch, so that travis will rebuild and hopefully pass, so that I can merge it?

Adminiuga:
@AndreasBomholtz I guess it is chick and the egg problem. I was thinking about adding another commit to setup.py, which you can omit when merging to install zigpy from github. I just need to figure out how. I've tried
```
    dependency_links=[
        'https://github.com/rcloran/pure-pcapy-3/archive/e289c7d7566306dc02d8f4eb30c0358b41f40f26.zip#egg=pure_pcapy3-1.0.1',
        'git+git://github.com/zigpy/zigpy.git@master#egg=zigpy',
    ],
```
but it still fails. Any ideas how to tell tox to install zigpy from github?

dmulcahey:
Why not just define the type in both and break the funky dependency... sometimes duplication is a better option. Ultimately a common library for the shared objects may make sense...

Adminiuga:
I was thinking the BroadcastAddress def to be reused in other radios as well. Unless we have any other ideas, we may have to go this route. @AndreasBomholtz  thoughts?

Adminiuga:
So I was able to make travis happy by modifying tox.ini to install zigpy dependency from github directly. I'm not sure about the best way to handle these changes as I don't have any experience with PyPi. 
Maybe we could publish new Zigpy to PyPi and then merge this PR but a855891458008196dd7b4cb7905609baacd102d1 commit and bump up zigpy dependency?

Adminiuga:
@AndreasBomholtz You rock! Thank you very much for your help and Merry Christmas!

AndreasBomholtz:
Thank you, and likewise :)

PR #157, opened 2019-04-06
Form

mtdcr:
Fix parameter -E (--extended-pan-id) and add parameter -N (--network-key) to bellows form command.

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -140,7 +140,7 @@ class ControllerApplication(zigpy.application.ControllerApplication):
             self._reset_task.cancel()
         self._ezsp.close()
 
-    async def form_network(self, channel=15, pan_id=None, extended_pan_id=[0] * 8):
+    async def form_network(self, channel=15, pan_id=None, extended_pan_id=[0] * 8, network_key=None):
         channel = t.uint8_t(channel)
 
         if pan_id is None:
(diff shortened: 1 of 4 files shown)


bellows/zigbee/application.py
```diff
@@ -140,17 +140,21 @@ def is_controller_running(self):
...
         if pan_id is None:
             pan_id = t.uint16_t.from_bytes(os.urandom(2), 'little')
         pan_id = t.uint16_t(pan_id)
 
-        if extended_pan_id is None:
-            extended_pan_id = t.fixed_list(8, t.uint8_t)([t.uint8_t(0)] * 8)
+        extended_pan_id = t.fixed_list(8, t.uint8_t)([x for x in extended_pan_id])
 
-        initial_security_state = bellows.zigbee.util.zha_security(controller=True)
+        if network_key is not None:
```
Adminiuga: why overwrite the supplied `network_key`?

Adminiuga:
Can you please add a test for https://coveralls.io/builds/22642248/source?filename=bellows/zigbee/application.py#L153 code? To keep coveralls happy

bellows/zigbee/application.py
mtdcr: In order to convert user input to the required data type for zha_security().

mtdcr:
I guess I would have added a test, if I knew how to do it.

bellows/zigbee/application.py
Adminiuga: Nevermind, was checking changes on the phone and confused it with 2nd change creating random key. All is good.

Adminiuga:
Implemented in #257

PR #169, opened 2019-06-20
Fix bug causing NCP to not resubscribe to multicast after restart

bballwiz5:
Since the list of multicast groups was not reset during initialization,
the controller would not resubscribe to multicast groups
it was a part of until the whole system was rebooted.


Adminiuga:
Good catch!

PR #171, opened 2019-06-22
RFC: Optionally Run Uart Protocol on a separate thread

bballwiz5:
Asyncio is designed for high number of connections and high event throughput. It is not designed for low latency. This causes an issue where bellows is not able to ack some packets quickly enough.

There are two main ways the protocol can be delayed:
1. High CPU usage in this process blocking the GIL
2. Long running or numerous async tasks sharing the event loop

This PR only deals with the second one. At least in my setup, it resolved all the NCP reset errors (pi 3b+). Resolving the first would require moving to a separate process, which involves a lot of marshaling of data.

I tried to find if anyone has talked about separating bellows into a different thread or process. My searching turned up references to running bellows in a separate thread, but no actual discussion of it. 
There is increased complexity in mixing asyncio and multiple event loops on separate threads. Hopefully since the uart protocol is the only part that has the latency requirement and has a much more restricted API, that complexity is manageable.

This is a true RFC, I just wanted to get something working to spark the discussion.

@Adminiuga thoughts?

### DIFF
--- a/bellows/__init__.py
+++ b/bellows/__init__.py
@@ -1,0 +1,5 @@
+MAJOR_VERSION = 0
+MINOR_VERSION = 7
+PATCH_VERSION = '3'
+__short_version__ = '{}.{}'.format(MAJOR_VERSION, MINOR_VERSION)
+__version__ = '{}.{}'.format(__short_version__, PATCH_VERSION)
--- a/bellows/__init__.py
+++ b/bellows/__init__.py
@@ -1,5 +1,5 @@
 MAJOR_VERSION = 0
-MINOR_VERSION = 7
-PATCH_VERSION = '3'
+MINOR_VERSION = 8
+PATCH_VERSION = '0'
 __short_version__ = '{}.{}'.format(MAJOR_VERSION, MINOR_VERSION)
 __version__ = '{}.{}'.format(__short_version__, PATCH_VERSION)
--- a/bellows/__init__.py
+++ b/bellows/__init__.py
@@ -1,5 +1,5 @@
 MAJOR_VERSION = 0
 MINOR_VERSION = 8
-PATCH_VERSION = '2'
+PATCH_VERSION = '2.dev0'
 __short_version__ = '{}.{}'.format(MAJOR_VERSION, MINOR_VERSION)
 __version__ = '{}.{}'.format(__short_version__, PATCH_VERSION)
--- a/bellows/__init__.py
+++ b/bellows/__init__.py
@@ -1,5 +1,5 @@
 MAJOR_VERSION = 0
 MINOR_VERSION = 8
-PATCH_VERSION = '0'
+PATCH_VERSION = '1'
 __short_version__ = '{}.{}'.format(MAJOR_VERSION, MINOR_VERSION)
 __version__ = '{}.{}'.format(__short_version__, PATCH_VERSION)
--- a/bellows/__init__.py
+++ b/bellows/__init__.py
@@ -1,5 +1,5 @@
 MAJOR_VERSION = 0
 MINOR_VERSION = 8
-PATCH_VERSION = '1'
+PATCH_VERSION = '2'
 __short_version__ = '{}.{}'.format(MAJOR_VERSION, MINOR_VERSION)
 __version__ = '{}.{}'.format(__short_version__, PATCH_VERSION)
--- a/bellows/commands.py
+++ b/bellows/commands.py
@@ -97,16 +97,9 @@ COMMANDS = {
     'pollHandler': (0x44, (), (t.EmberNodeId, )),
     'incomingSenderEui64Handler': (0x62, (), (t.EmberEUI64, )),
     'incomingMessageHandler': (0x45, (), (t.EmberIncomingMessageType, t.EmberApsFrame, t.uint8_t, t.int8s, t.EmberNodeId, t.uint8_t, t.uint8_t, t.LVBytes)),
-    # TODO: LVBytes in incomingRouteRecordHandler is really a list of
-    #       uint16_t - decoding this frame will be incorrect, since the length
-    #       refers to the number of items in the list
-    'incomingRouteRecordHandler': (0x59, (), (t.EmberNodeId, t.EmberEUI64, t.uint8_t, t.int8s, t.LVBytes)),
+    'incomingRouteRecordHandler': (0x59, (), (t.EmberNodeId, t.EmberEUI64, t.uint8_t, t.int8s, t.LVList(t.EmberNodeId))),
     'changeSourceRouteHandler': (0xC4, (), (t.EmberNodeId, t.EmberNodeId, t.Bool)),
-    # TODO: LVBytes in setSourceRoute is really a list of uint16_t,
-    #       serializing the frame will produce something the NCP will fail
-    #       on, since the NCP expects the length to be number of items in the
-    #       list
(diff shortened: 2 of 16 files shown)


Adminiuga:
Wow! this is impressive.
I concur, UART portion is the only one sensitive to latency and running it in a separate thread makes sense as it has quite restricted interaction with the upper layers. 

Out of curiosity, how large is your `home-assistant_v2.db` file?

bballwiz5:
At the moment 500MB. Though it only covers a small portion of my house. I'm about to migrate from Wink which runs the rest of the house.

dmulcahey:
I think this is a good idea.

Adminiuga:
Per EZSP ASH protocol specs, I'd say initial parameters of NCP waiting for an ACK are quite lax
image

Starting a 1.6s. I wonder if `bellows` always trying to send the ACK immediately shrinks that window (the window is a moving average across last ACKs), so when CPU is under the load it doesn't have cycles to sends the ACKs that quickly anymore, causing it to violate the ACK timeout.  But never had the time to test it. 

I guess I'll have to move my prod back to Pi2 and run it for a while, as I never could reproduce it in the dev environment. In production, I only could get it to enter the failed state only if my recorder DB grew over 500-700mb. Right now it is about 150mb and I don't get any restarts. On the same hardware & config, but with DB in 1.2GB range I was getting a restart about every few hours. 

I won't be able to get this into 0.95, but going to run it for a week and shoot for 0.96. But we'll have to fix lint and tests.

bballwiz5:
Interesting that it uses a sliding window. I guess we are penalized for being too fast too often :). I've been running this branch for about a week now with no more resets. Let me know if you encounter any.

Tests and linting are fixed along with a bug where it would hang on restart.

Adminiuga:
just wanted to update everyone that it is still on my radar. I moved my production back to Pi 2 just to establish the baseline. I had only one reset in 36 hours and incidentally the reset happened about after 24 hours the HA was started. Going to upgrade to 0.95 so will reset it one more time.

bballwiz5:
Thanks @Adminiuga. Since it seems everyone agrees this is a good idea, I cleaned up the code and made it a bit more testable and maintainable by moving all the threading and proxying to a separate module.

Adminiuga:
so I'm still trying to establish the baseline, the problem is, that my production running on Pi2 (and let me tell you, it is much slower than Pi3+) has not had failure resets for the last three days. I'll leave it for a few more days and then resort to `iozone` to overload the system.

BTW, I suspect this could be related to the fix in history component as previously running my prod on this Pi2 was yielding at least one reset per 36 hours.

Adminiuga:
status update:
About a week ago my recorder DB went over 512MB and somewhere above that line, resets started poping up in the log. I've switched to your branch, but HA was just unusable (had nothing to do with your branch) commands would take about 2-3 sec to take effect, SD card was under constant IO, per IO top there were about 600KB/s sustained writes and HA was top IO hog hovering in 99%-100% zone. 

Turns out, somehow the recorder DB got corrupted. Don't know why or how, but previously on this hardware I also was seeing problems when DB was over 512MB. Maybe there's a bad block on SD card, dunno. I've tried repairing the database and thought I repaired it, but in reality it wasn't logging anything in it, so I had to kill recorder DB and start from scratch. in 5 days it is about 285MB atm. I did not have restarts causes by ACK timeouts, but to be fair I didn't have restarts with upstream bellows in similar conditions. 

One thing I've noticed, I cannot shutdown HA anymore. It keeps about 4 threads running and won't shutdown even after 20min waiting. Haven't tested this in other environments though

Adminiuga:
just tested in my dev and it would also not shutdown HA properly. Are we closing/joining the thread on serial close?

Adminiuga:
it looks like HA is patching `asyncio.Task.all_tasks()` 
image

`all_tasks()` returns `IgnoreCalls` object and it isn't iterable, so https://github.com/bballwiz5/bellows/blob/56efeefda784e1299204444e2ac69089345a6892/bellows/thread.py#L18 never completes, just throws.

Adminiuga:
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/monkey_patch.py

Adminiuga:
Removing that monkey patching fixes the issue. The million $ question whether that patching is still needed.

Adminiuga:
maybe we should leave it as is. HA 0.97 will support Python 3.7+ and shouldn't patch `all_tasks()`

bballwiz5:
That's an odd patch. Well, its easy enough to remove the dependency of `all_tasks()` and explicitly stop the thread when the serial connection is done

Interesting that you don't see the resets as often as I did before I started running with this code. I wonder if it has to do with the number of Lightify bulbs that I have which all report a power sensor that is force updated at the same time for all the bulbs (useless sensor anyway). That seemed to coincide with the bellows resets. I wonder if that would cause a bunch of IO in recorder.

Since I've been running with this code, I haven't seen any resets.

bballwiz5:
I'm fine with either waiting for HA to remove the monkeypatch or using the version that doesn't rely on `all_tasks`. Whichever you prefer. 

I don't know why the `application` unit tests are suddenly failing. They pass locally and I didn't touch that part of the code

Adminiuga:
> interesting that you don't see the resets as often as I did before I started running with this code. I wonder if it has to do with the number of Lightify bulbs that I have which all report a power sensor

How many do you have. I currently count about 70ish devices, but only 15 power metering (1794/0x0702 cluster id) sensors. I'm also running on Pi2 (half of Pi 3+ CPU cycles), cause Pi3 is on hassio and I couldn't figure out how to maintain logs there.

Interesting that you see those on updates, we are polling 5 devices max concurrently.

Adminiuga:
> I don't know why the application unit tests are suddenly failing

Try killing local `.tox` folder. Although I don't think there were significant changes in zigpy

Adminiuga:
it's because of https://github.com/zigpy/zigpy/commit/13b12a2993d3f61767951b104f1180d7a692381b 
I need to fix it.

Adminiuga:
rebase or pick 6de019339af7442b4463a26af47eaeb259973ec6

Adminiuga:
I had no problems running this branch so far. BTW updated to Python3.7 and the issue with HA restarts was solved.
The only problem I'm having with Pi2 is heavy IO and therefore lag once recorder DB grows beyond 550MB. Had to kill my DB, cause starting HA was way to slow. 

Does threading affect any bellows CLI ?

bballwiz5:
It shouldn't affect bellows CLI. The CLI will just use the threaded version. There shouldn't be a problem with using the threaded version. The thread will still stop when the connection is shut down.

bballwiz5:
I reverted my system to not use this patch and started getting resets again. The new recorder code definitely helps quite a bit, my database is now ~15MB and I don't think the recorder is the cause anymore. However, anytime my pi does high IO, I end up getting an NCP reset (for instance while updating HA through hassio). I think this PR would still be helpful, even if the sources of high IO are being reduced, I don't think they will ever be eliminated.

Adminiuga:
Thank you!
I ran your branch for a few weeks without issues. No resets even when recorder was going bananas doing excessive disk IO when db grows over 512MB, essentially making the system hardly usable at all.

Adminiuga:
@bballwiz5 the tests are failing randomly. eg https://travis-ci.org/zigpy/bellows/builds/562687698 and https://travis-ci.org/zigpy/bellows/builds/562687897

And sometimes they randomly pass. Any ideas what is going on?

PR #212, opened 2020-02-02
Update README.md to mention a few related projects for reference

Gamester17:
Update README.md to describe zigpy and also mention these related projects for reference

zha-map https://github.com/zha-ng/zha-map
zha-network-visualization-card https://github.com/dmulcahey/zha-network-visualization-card
zha-network-card https://github.com/dmulcahey/zha-network-card


Adminiuga:
Thank you for  your contribution. but I don't think this belongs here. We probably need to wait until ZHA/HomeAssistant documentation structure gets sorted out or start a dedicate ZHA page.

PR #217, opened 2020-03-07
Remove EZSP join handler

dmulcahey:
This PR removes the EZSP join handler. All devices were triggering handle_join 2x because device announce was also triggering the join handler.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -202,12 +202,6 @@ def test_dup_send_success(app, aps, ieee):
     assert req.result.set_result.call_count == 1
 
 
-def test_join_handler(app, ieee):
-    # Calls device.initialize, leaks a task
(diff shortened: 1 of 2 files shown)


tests/test_application.py
```diff
@@ -202,12 +202,6 @@ def test_dup_send_success(app, aps, ieee):
     assert req.result.set_result.call_count == 1
 
 
-def test_join_handler(app, ieee):
-    # Calls device.initialize, leaks a task
-    app.ezsp_callback_handler("trustCenterJoinHandler", [1, ieee, None, None, None])
-    assert ieee in app.devices
```
Adminiuga: Let's leave this test in, and assert that `app.handle_join.call_count == 0`

PR #226, opened 2020-03-29
ezsp: fix incorrect ezsp_version reference

jbott:
Similar fix, looks like this got missed in refactoring.


Adminiuga:
Doh. Good catch.

PR #245, opened 2020-04-21
Update README.md with port configuration hints and suggestions

Hedda:
Update README.md with port configuration hints and suggestions. Baud rate suggestion copied from https://github.com/zsmartsystems/com.zsmartsystems.zigbee so credits go to @cdjackson


Adminiuga:
EM3588 is a Silicon Labs chip, i don't think it is present on Ti boards.

Hedda:
Copy and paste mistake

PR #247, opened 2020-04-21
Update README.md with port configuration hints and suggestions

Hedda:
Update README.md with port configuration hints and suggestions. Baud rate suggestion copied from https://github.com/zsmartsystems/com.zsmartsystems.zigbee so credits go to @cdjackson

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ $ bellows zcl 00:0d:6f:00:05:7d:2d:34 1 1026 read_attribute 0
 0=1806
 ```
 
+## Port configuration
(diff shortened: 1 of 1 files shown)


Adminiuga:
`bellows` does not support `auto` port detection and probably would never will, just because of the variety of USB-to-Serial adapters out there. 
Some sort of "auto-detection" would/should be implemented in ZHA, asking user to pick a port and then probing radios on that port based on radio defaults. Gonna be challenging just because there's a variety of zigpy based radios.

I'd remove wording about `auto` but leave the EM3588 and part where there are mostly two types of images with 115200  and 57600baud rattes

Hedda:
OK cool

PR #281, opened 2020-07-29
Add EZSP protocol version 5

Adminiuga:
EZSP protocol version 5 types and commands.
Version 5 frame handling.


MattWestb:
Great work !!
Near 4500 changes together with the first general commit so not a piece of cake and hope the code its working well so you don't need spending your time correcting bugs and can using your time for the higher versions.

PR #295, opened 2020-08-11
Export/Import NCP Trust Center parameters

Adminiuga:
Theoretically this allows backing up and restoring NCP state between the hardware version. 
Please note this is highly experimental. The restoration doesn't restore NCP children and relies on children just to find a new parent. 
Another warning: if this is done between different hardwares, the EUI64 is going to be different on the new hardware meaning the binding tables on all the devices are going to be wrong. 
You either have to reconfigure all the devices -- recommended so the bindings are updated or alternatively you could override the EUI64 on the new hardware, essentially producing a clone. **Be very careful** if you decide to go this route. You can change the EUI64 only once and once you set it it is impossible to change it without a specialized hardware (SWD programmer).

To export TC config, see `bellows backup --help` usually it is just `bellows backup > your-backup-file.txt`. The backup contains your network key, so you probably should keep that file secure.
To import, see `bellows restore --backup-file your-backup-file.txt`

### DIFF
--- a/bellows/cli/__init__.py
+++ b/bellows/cli/__init__.py
@@ -1,2 +1,2 @@
 # flake8: noqa
-from . import application, dump, ncp, network
+from . import application, backup, dump, ncp, network
--- a/bellows/cli/backup.py
+++ b/bellows/cli/backup.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import logging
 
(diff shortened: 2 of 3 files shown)


Hedda:
PR for a suggested update to README.md with information about this new feature -> https://github.com/zigpy/bellows/pull/356

PR #306, opened 2020-08-16
Change the default TC policy for EZSPv8 to allow unsecure rejoins

Adminiuga:
Change the default TC policy for EZSPv8 to allow unsecure rejoins.


adriweb:
What's the reason behind this? Wouldn't [this](https://know.bishopfox.com/hubfs/blackhatusa2019-zigbeehacking-smarterhomeinvasionwithzigdiggity-08aug2019-slides-190822171804.pdf) be a security problem?

Adminiuga:
it was reverted in https://github.com/zigpy/bellows/pull/312

It is a security problem, however some users reported their ZB3 devices (Ecosmar bulbs) dropping off after failing an unsecure rejoin, so the default may be changed back to allow unsecure rejoins. 

What EZSP version are you running and do you experience any problems? I need more feedback from different use cases

adriweb:
Ah, I didn't see #312, thanks :)

I'm no a user of bellows myself currently, but I like following its development as I may use it one day!

PR #316, opened 2020-08-28
Use Hashed TC link key

Adminiuga:
Implement https://www.silabs.com/community/wireless/zigbee-and-thread/knowledge-base.entry.html/2018/06/18/hashed_link_keys-Ltmt

As side effect add well known TC global link key as "transient key" to allow interoperability with HA12 devices.

### DIFF
--- a/bellows/cli/backup.py
+++ b/bellows/cli/backup.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+import os
 
 import bellows.types as t
 import click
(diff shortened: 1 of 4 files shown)


MattWestb:
First fast test before going to sleep all 9 devices in.
Little strange is that the IKEA bulbs was paring itself then one was rebooted (all devices was not reseted and was paired in tasmota so knowing the NCP address but hopefully not PAN, E-PAN and NWK).
ZB3 paring looks being very fast !!!
Xiaomi is doing xiaomi things and taking longer.
As normal great work done !!!

PR #359, opened 2020-10-17
Fix tests package exclusion

scop:


Adminiuga:
Does exclusion mask of `packages=find_packages(exclude=["tests", "tests.*"])` introduced in https://github.com/zigpy/bellows/pull/368 works for you?

scop:
That works too, yes.

PR #378, opened 2020-12-08
Update setup.py adding "pyusb>=1.1.0" requirement

Hedda:
Update setup.py adding "pyusb>=1.1.0" requirement as probably does not hurt have latest pyusb / libusb installed when trying to add USB adapters.

Reference https://github.com/zigpy/zigpy-zigate/issues/71 and https://github.com/home-assistant/home-assistant.io/pull/15829


Adminiuga:
Bellows does not use `pyusb` nor pyusb is going to update the libusb. So it only should include dependencies on the packages it really uses.

PR #380, opened 2020-12-11
Bump up dependencies on pyserial and pyserial-asyncio

Hedda:
Bump up dependencies on pyserial and pyserial-asyncio for projects other projects that use this library but are not Home Assistant.

Adminiuga did the same dependency version bump for Home Assistant in https://github.com/home-assistant/core/pull/44089

pyserial-asyncio depends on pyserial and this bumps pyserial-asyncio to 0.5 version release and pyserial to 3.5 version release.

https://github.com/pyserial/pyserial-asyncio/releases/tag/v0.5

https://github.com/pyserial/pyserial/releases/tag/v3.5


Adminiuga:
Per se bellows works with older versions too, so it is better to pin specific version at applications relying on bellows.

PR #384, opened 2021-01-03
cli: join network fails because of typo

basilfx:
I'm new to bellows, and I am testing it via the CLI. I noticed that `network.extendedPanId` must be used instead of `network.ExtendedPanId` for a `bellows join` command.

I still don't have any success (next error: `TypeError: zha_security() missing 1 required positional argument: 'config'`), but at least I'm one step further with this fix.


Adminiuga (approved):

Adminiuga:
Good catch

PR #390, opened 2021-01-22
Update README.md with Bitron adapter and port configuration issues

Hedda:
Add the Bitron Video/Smabit BV AV2010/10 USB-Stick adapter and also update tips on discovery issues for port configuration.

Part of the port config guide borrowed from https://www.openhab.org/addons/bindings/zigbee/#ember-ezsp-ncp-coordinator


Hedda:
Note that should also be added to bellows as the plan is to remove the whole list, inc. it, from zigpy README via https://github.com/zigpy/zigpy/pull/705

Downstream project(s) also want such information to be provided by upstream as per https://github.com/home-assistant/home-assistant.io/pull/17089

Hedda:
Closing as split and replaced by https://github.com/zigpy/bellows/pull/412 and https://github.com/zigpy/bellows/pull/411

PR #395, opened 2021-02-18
Update README.md

MattWestb:
SPI protocols over a serial interface is not implanted in bellows (wot i knowing)
Port configuration: 
- serial port for no USB adapters
- software flow control
- auto discover modules with 57600 baud.

Tuya ZBGW  network config.
 EZSP protocol v8  is not backward compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ zigpy requires a robust connection between the zigpy radio library and the seria
 
 ## Firmware requirement
 
-bellows requires that the Zigbee adapter/board/module is pre-flashed/flashed with compatible firmware with EmberZNet PRO Zigbee Stack that uses the standard Silicon Labs EZSP (EmberZNet Serial Protocol) APIs for ASH or SPI protocols over a serial interface.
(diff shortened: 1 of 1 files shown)


tube0013:
Maybe not define the port for the socket connection as that can be configured to anything on the software on the serial to network bridge.

MattWestb:
Then i think also making it "neutral" and not branded is better then its coming more network radios (then JFK have released all parts).

MattWestb:
@tube0013 Is you planing using only 115k2 as port speed so it can still being in the example or shall i also making  it "neutral" ?

Its also possible configure the NCP with hardware flow control and Socat is using it on the NCP side and the the TCP flow control to ZHA but then you need implanting the hardware lines on the PCB but its one small "enchantment" in the design (tuya is using it on there ZBGW and its looks working OK).

MattWestb:
Have making the network adapter config "neutral" for future network adapter being added to bellows.

I see one urgent need of re construction of both Zigpy and the Bellows readme.md and moving information to Zigpy Wiki so not having so much "low level information" and duplicate information that is taking place and making things more confusing for new inexperience users.
One example is all lines with baud rate and flow control, firmware needed and recommended (more lines / adapter) can being moved to  the Wiki and being linked from the "main" page (as "see wiki for detailed information" as i have making with the firmware matrix in the wiki) for keeping more clean and being easier to understanding.
But Im not the right person doing it then im writing English is im spiking :-((

Adminiuga (approved):

PR #397, opened 2021-02-28
Increase maximums to support 'pro' devices

dmulcahey:
Increase the max for children and neighbors to support 'pro' type devices that can handle the settings. 
Remove the default for MAX_END_DEVICE_CHILDREN for V8 EZSP. Inherit the default from NCP

### DIFF
--- a/bellows/ezsp/v8/config.py
+++ b/bellows/ezsp/v8/config.py
@@ -22,7 +22,7 @@ EZSP_SCHEMA = {
     # The maximum number of router neighbors the stack can keep track of. A neighbor
     # is a node within radio range
     vol.Optional(EzspConfigId.CONFIG_NEIGHBOR_TABLE_SIZE.name): vol.All(
-        int, vol.Range(min=8, max=16)
+        int, vol.Range(min=8, max=26)
     ),
     #
     # The maximum number of APS retried messages the stack can be transmitting at
@@ -80,7 +80,7 @@ EZSP_SCHEMA = {
     #
     # The maximum number of end device children that a router will support
     vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name, default=32): vol.All(
-        int, vol.Range(min=0, max=32)
+        int, vol.Range(min=0, max=64)
     ),
     #
     # The maximum amount of time that the MAC will hold a message for indirect
--- a/bellows/ezsp/v8/config.py
+++ b/bellows/ezsp/v8/config.py
@@ -79,7 +79,7 @@ EZSP_SCHEMA = {
     ),
     #
     # The maximum number of end device children that a router will support
-    vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name, default=32): vol.All(
+    vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name): vol.All(
         int, vol.Range(min=0, max=64)
     ),
     #


bellows/ezsp/v8/config.py
```diff
@@ -22,7 +22,7 @@
     # The maximum number of router neighbors the stack can keep track of. A neighbor
     # is a node within radio range
     vol.Optional(EzspConfigId.CONFIG_NEIGHBOR_TABLE_SIZE.name): vol.All(
```
tube0013, line 24: should this get a default added for 16?

Adminiuga: Nope. Without setting a default it would use the value the NCP image was compiled with. 
And in other words I suspect that for v8 version and newer maybe it is better not to set any values at all? E.g. if the ncp image was compily with 16 tc keys slots, you won't be able to set a higher value. Essentially you only may decrease values, but not increase them. But I'm not positive this applies to all values.

bellows/ezsp/v8/config.py
```diff
@@ -80,7 +80,7 @@
     #
     # The maximum number of end device children that a router will support
     vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name, default=32): vol.All(
-        int, vol.Range(min=0, max=32)
+        int, vol.Range(min=0, max=64)
```
Adminiuga, line 83: another approach, we increase the max in the schema and remove the default. 
Just need to confirm that on NCP compiled with 16 children, you can increase it to 32 or not.

Adminiuga: this is what I get trying to set children to 33 on 6.7.8.0 build
```
debug: Application frame 0 (version) received: b'08026067'
debug: EZSP Stack Type: 2, Stack Version: 6760, Protocol version: 8
debug: Send command setConfigurationValue: (<EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN: 17>, 33)
debug: Sending: b'224321a9072a0493595a847e'
debug: Data frame: b'2343a1a9072a2302c27e'
debug: Sending: b'83401b7e'
debug: Application frame 83 (setConfigurationValue) received: b'36'
[<EzspStatus.ERROR_INVALID_VALUE: 54>]
debug: Closed serial connection
```

So my question: if you compile the image with defaults set to 16, would you be able setting this to 32? @tube0013 wanna give it a try? If compiled image with value 16 does not let you set 32 children, then we just should remove the default from the schema. Would save an extra step and would be enabled automatically with your firmare

tube0013: I will test tonight. how would I tell if it takes the updated value?

I'm using 6.9.1 as I can't find the download for 6.7.8 any longer in the SDK manager in v5. It seems the fixes in 6.7.8 are in 6.9.1.

Adminiuga: you can do `bellows -d /dev/ttyUSBx config --all` it would print the ~~current~~ default, out of the box settings

tube0013: compiled FW with 16 child. config set for 32....

```2021-03-01 22:43:31 WARNING (MainThread) [bellows.ezsp.protocol] Couldn't set EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN=32 configuration value: EzspStatus.ERROR_INVALID_VALUE
```

Adminiuga: Means we are better off not setting this value at all? As we can only lower it?

tube0013: > you can do `bellows -d /dev/ttyUSBx config --all` it would print the ~current~ default, out of the box settings

this is with children set to 64 in fw.

`bellows -d socket://192.168.1.91:6638 config --all
NOTE: Configuration changes do not persist across resets
CONFIG_PACKET_BUFFER_COUNT=254
CONFIG_NEIGHBOR_TABLE_SIZE=26
CONFIG_APS_UNICAST_MESSAGE_COUNT=30
CONFIG_BINDING_TABLE_SIZE=32
CONFIG_ADDRESS_TABLE_SIZE=32
CONFIG_MULTICAST_TABLE_SIZE=8
CONFIG_ROUTE_TABLE_SIZE=16
CONFIG_DISCOVERY_TABLE_SIZE=8
CONFIG_STACK_PROFILE=0
CONFIG_SECURITY_LEVEL=5
CONFIG_MAX_HOPS=30
CONFIG_MAX_END_DEVICE_CHILDREN=64
CONFIG_INDIRECT_TRANSMISSION_TIMEOUT=3000
CONFIG_END_DEVICE_POLL_TIMEOUT=8
CONFIG_TX_POWER_MODE=0
CONFIG_DISABLE_RELAY=0
CONFIG_TRUST_CENTER_ADDRESS_CACHE_SIZE=0
CONFIG_SOURCE_ROUTE_TABLE_SIZE=200
CONFIG_FRAGMENT_WINDOW_SIZE=1
CONFIG_FRAGMENT_DELAY_MS=0
CONFIG_KEY_TABLE_SIZE=12
CONFIG_APS_ACK_TIMEOUT=1600
CONFIG_ACTIVE_SCAN_DURATION=3
CONFIG_END_DEVICE_BIND_TIMEOUT=60
CONFIG_PAN_ID_CONFLICT_REPORT_THRESHOLD=2
CONFIG_REQUEST_KEY_TIMEOUT=0
CONFIG_CERTIFICATE_TABLE_SIZE=0
CONFIG_APPLICATION_ZDO_FLAGS=0
CONFIG_BROADCAST_TABLE_SIZE=15
CONFIG_MAC_FILTER_TABLE_SIZE=0
CONFIG_SUPPORTED_NETWORKS=1
CONFIG_SEND_MULTICASTS_TO_SLEEPY_ADDRESS=0
CONFIG_ZLL_GROUP_ADDRESSES=1
CONFIG_ZLL_RSSI_THRESHOLD=128
CONFIG_MTORR_FLOW_CONTROL=1
CONFIG_RETRY_QUEUE_SIZE=16
CONFIG_NEW_BROADCAST_ENTRY_THRESHOLD=9
CONFIG_TRANSIENT_KEY_TIMEOUT_S=300
CONFIG_BROADCAST_MIN_ACKS_NEEDED=255
CONFIG_TC_REJOINS_USING_WELL_KNOWN_KEY_TIMEOUT_S=300
CONFIG_CTUNE_VALUE=282
CONFIG_ASSUME_TC_CONCENTRATOR_TYPE=1`

bellows/ezsp/v8/config.py
```diff
@@ -80,7 +80,7 @@
     #
     # The maximum number of end device children that a router will support
     vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name, default=32): vol.All(
```
Adminiuga, line 82:
```suggestion
    vol.Optional(EzspConfigId.CONFIG_MAX_END_DEVICE_CHILDREN.name): vol.All(
```

Adminiuga (approved):

MattWestb:
One question is the "PRO" upgrade working as expected = extending the neighbors and clients is the firmware is having it in there config  on V8 EZSP ?

I have putting one "EZSO PRO" column in the coordinate wiki firmware for devices that have v8 firmware but im still missing tube013s firmware for his "EZSP Black PRO".

PR #428, opened 2021-07-22
Automatically fix the frame FCS in `bellows dump`

puddly:
https://github.com/zigpy/zigpy-cli/pull/1#issuecomment-884638373

The FCS is validated by EmberZNet so the frames are valid, only the checksum is wrong.

### DIFF
--- a/bellows/cli/dump.py
+++ b/bellows/cli/dump.py
@@ -80,9 +80,8 @@ async def _dump(ctx, channel, outfile):
 
             # Later releases of EmberZNet incorrectly use a static FCS
             fcs = data[-2:]
-            computed_fcs = ieee_15_4_fcs(data[0:-2])
-
-            if s.ezsp_version == 8 and fcs == b"\x0F\x00" and fcs != computed_fcs:
+            if s.ezsp_version == 8 and fcs == b"\x0F\x00":
+                computed_fcs = ieee_15_4_fcs(data[0:-2])
                 LOGGER.debug("Fixing FCS (expected %s, got %s)", computed_fcs, fcs)
                 data = data[0:-2] + computed_fcs
 
--- a/bellows/cli/dump.py
+++ b/bellows/cli/dump.py
@@ -77,11 +77,14 @@ async def _dump(ctx, channel, outfile):
     def cb(frame_name, response):
         if frame_name == "mfglibRxHandler":
             data = response[2]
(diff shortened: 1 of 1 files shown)


bellows/cli/dump.py
```diff
@@ -60,6 +77,12 @@ async def _dump(ctx, channel, outfile):
     def cb(frame_name, response):
         if frame_name == "mfglibRxHandler":
             data = response[2]
+            fcs = ieee_15_4_fcs(data[0:-2])
+
+            if data[-2:] != fcs:
+                LOGGER.warning("Fixing frame FCS (expected %s, got %s)", fcs, data[-2:])
```
Adminiuga, line 83: That's going to be a lot of logging. Make it debug instead.

bellows/cli/dump.py
```diff
@@ -60,6 +77,12 @@ async def _dump(ctx, channel, outfile):
     def cb(frame_name, response):
         if frame_name == "mfglibRxHandler":
             data = response[2]
+            fcs = ieee_15_4_fcs(data[0:-2])
+
+            if data[-2:] != fcs:
```
Adminiuga, line 82: I would not do it unconditionally, but only if ezsp version == 8 and last two bytes of the frame that static value

Adminiuga (commented):
Thank for the fix

bellows/cli/dump.py
```diff
@@ -60,6 +77,15 @@ async def _dump(ctx, channel, outfile):
...
         if frame_name == "mfglibRxHandler":
             data = response[2]
+
+            # Later releases of EmberZNet incorrectly use a static FCS
+            fcs = data[-2:]
+            computed_fcs = ieee_15_4_fcs(data[0:-2])
+
+            if s.ezsp_version == 8 and fcs == b"\x0F\x00" and fcs != computed_fcs:
+                LOGGER.debug("Fixing FCS (expected %s, got %s)", computed_fcs, fcs)
+                data = data[0:-2] + computed_fcs
```
Adminiuga, line 87:
```suggestion
            if s.ezsp_version == 8 and fcs == b"\x0F\x00":
                computed_fcs = ieee_15_4_fcs(data[0:-2])
                LOGGER.debug("Fixing FCS (expected %s, got %s)", computed_fcs, fcs)
                data = data[0:-2] + computed_fcs
```

Adminiuga (approved):
Thank you!

PR #441, opened 2021-11-17
Set specific manufacturer ID based on the joining device IEEE address

Adminiuga:
Set specific manufacturer id reported in node descriptor, based on the joining device IEEE address.


dmulcahey:
I say this is good to go and we can deal with any fallout afterwards w/ quirks changes for Xiaomi devices if any are acting differently like the T1 motion sensor. I already have a quirk for that working.

PR #447, opened 2022-01-27
Skip get_board_info() in case EZSP Radio doesn't support it

agners:
Handle exceptions in case EZSP Radio doesn't support the getMfgToken
token command. This is required to make ZHA work with the EmberZNet
Zigbee stack running on Linux (zigbeed).

See also discussion: https://github.com/zigpy/zigpy/discussions/894

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -145,10 +145,13 @@ class ControllerApplication(zigpy.application.ControllerApplication):
             output_clusters=[zigpy.zcl.clusters.security.IasZone.cluster_id]
         )
 
-        brd_manuf, brd_name, version = await self._ezsp.get_board_info()
-        LOGGER.info("EZSP Radio manufacturer: %s", brd_manuf)
-        LOGGER.info("EZSP Radio board name: %s", brd_name)
-        LOGGER.info("EmberZNet version: %s", version)
+        try:
+            brd_manuf, brd_name, version = await self._ezsp.get_board_info()
+            LOGGER.info("EZSP Radio manufacturer: %s", brd_manuf)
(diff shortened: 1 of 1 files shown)


agners:
FWIW, `zigbeed` from Gecko SDK v4.0.1 has this fixed now. The command returns with this information :see_no_evil: 

```
2022-01-29 22:34:44 INFO (MainThread) [bellows.zigbee.application] EZSP Radio manufacturer: b'\x00\x00\xeby\x00\x11\x01\x7f\x90)\x8e\r,\x00\xb9\xb0'
2022-01-29 22:34:44 INFO (MainThread) [bellows.zigbee.application] EZSP Radio board name:                                        
2022-01-29 22:34:44 INFO (MainThread) [bellows.zigbee.application] EmberZNet version: 7.0.1.0 build 0
```

Adminiuga:
whelp, not really helpful :facepalm: 

Are these tokens EMberZNet Specific and are they still applicable in case of an RCP firmware running on SOC?

agners:
When creating a RCP project, and looking into "Software components", it seems that the "Token Manager" is not installed. So I'd say it doesn't use those tokens.

Afaik, the RCP firmware comes from OpenThread side, it doesn't seem that much SiLabs/EmberZNet influenced.

> The RCP image comes in two flavors: multi-PAN (802.15.4 only), and multiprotocol (with Bluetooth added).
> 
> The multi-PAN 802-15.4-only RCP is based on the OpenThread 802.15.4 RCP with added multi-PAN and CPC support. It has a small flash footprint (~150K) and uses the Spinel protocol to serialize 802.15.4 commands. 

Spinel is also a OpenThread thing.

PR #451, opened 2022-03-11
Import CONF_NWK parameters to config

SylvainPer:
Allows network configuration at initialization without calling zigpy library

### DIFF
--- a/bellows/config/__init__.py
+++ b/bellows/config/__init__.py
@@ -4,6 +4,15 @@ from zigpy.config import (  # noqa: F401 pylint: disable=unused-import
     CONF_DEVICE_PATH,
     CONFIG_SCHEMA,
     SCHEMA_DEVICE,
+    CONF_NWK,
+    CONF_NWK_KEY,
+    CONF_NWK_PAN_ID,
+    CONF_NWK_CHANNEL,
+    CONF_NWK_CHANNELS,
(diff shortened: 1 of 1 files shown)


Adminiuga:
You don't need it. Use ApplicationController.Schema to validate network configuration schema, so it stays consistent between different libraries

SylvainPer:
It's not the validation that causes problem but the declaration.
Here the code I'm using that working with ZNP:

```python
import bellows.config as conf
config = {conf.CONF_DEVICE: {"path": serialPort, "baudrate": 115200}, conf.CONF_NWK: {}}
if set_extendedPanId != 0:
        config[conf.CONF_NWK][conf.CONF_NWK_EXTENDED_PAN_ID] = "%s" % (
            t.EUI64(t.uint64_t(set_extendedPanId).serialize())
        )
if set_channel != 0:
        config[conf.CONF_NWK][conf.CONF_NWK_CHANNEL] = set_channel
```

I can't do it the same way, what should I do ?

Adminiuga:
You are already using `"path"` and `"baudrate"`, so you could use string literals for the rest of the config, or import those from zigpy. Zigpy is installed anyway as part of bellows dependency.

SylvainPer:
Many thanks, it avoids several import (bellows and zigpy).

PR #454, opened 2022-03-18
Add stream subcommand for RF testing

agners:
The stream subcommmand uses the manufacturing library to transmit random
characters continously. This is useful for RF testing and approvals.


Adminiuga (approved):
Lgtm

PR #456, opened 2022-04-01
Add tone command for RF testing

agners:
From the mfglib description:
[...] start a TX tone that can be used to measure the PA output power,
phase noise, frequency offset and spurious emissions using a spectrum
analyzer. The TX tone is a continuous unmodulated transmission.


Adminiuga (approved):
LGTM!
Can you rebase against the dev, which should have the fix for black?

Adminiuga (approved):
Lgtm

PR #457, opened 2022-04-02
Fix black with latest click

agners:
Latest click broke black (see https://github.com/psf/black/issues/2964
for details). Use latest black to fix black in our CI environment. Also
add .pre-commit-config.yaml to the key of the cache.


Adminiuga (approved):
Thanks!

PR #459, opened 2022-05-10
Fix UART port close on RSTACK message during startup

agners:
When a RSTACK message is processed right after the UART has been opened,
it causes EZSP.enter_failed_state() getting called at a point where the
application callbacks are not registered yet. In that case the UART
will get closed and it won't get opened again. Bellows is stuck with a
closed transport.

Avoid this issue by not closing the port in case there is no application
callback registered yet.

Typically, it is unlikely that a RSTACK message arrives right when
the port gets opened (the race window is very narrow). However, with
hardware flow control opening the port leads to RTS signal to get
asserted which causes the radio to send pending messages, e.g. resets
caused by EmberZNet watchdog.

Note: With hardware flow control this is only the case if the tty "hupcl"
option is set. The option is set by default, but cleared by tools like
GNU screen. This option makes sure that the RTS signal is deasserted
while the port is closed. Pyserial/bellows does not change the state
of that option.

Fixes: https://github.com/home-assistant/core/issues/71575

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -224,7 +224,9 @@ class EZSP:
             self.close()
             self.handle_callback("_reset_controller_application", (error,))
         else:
-            LOGGER.info("NCP entered failed state. No application handler registered, ignoring...")
+            LOGGER.info(
+                "NCP entered failed state. No application handler registered, ignoring..."
+            )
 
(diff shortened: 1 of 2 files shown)


Adminiuga (approved):
LGTM

Adminiuga (approved):

PR #473, opened 2022-07-25
Handle null-terminated `MFG_` strings

puddly:
Backups were being created with invalid metadata:

```json
"ezsp": {
    "manufacturer": "Nabu Casa\u0000",
    "board": "SkyBlue v0.1\u0000",
    "version": "7.1.0.0 build 191",
    "stack_version": 9,
    "can_write_custom_eui64": false
}
```

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -267,19 +267,25 @@ class EZSP:
         """Return board info."""
 
         tokens = []
+
         for token in (t.EzspMfgTokenId.MFG_STRING, t.EzspMfgTokenId.MFG_BOARD_NAME):
-            LOGGER.debug("getting " "%s" " token", token.name)
-            (result,) = await self.getMfgToken(token)
+            (value,) = await self.getMfgToken(token)
+            LOGGER.debug("Read %s token: %s", token.name, value)
+
+            # Tokens are fixed-length and initially filled with \xFF
+            result = value.rstrip(b"\xFF").split(b"\x00", 1)[0]
+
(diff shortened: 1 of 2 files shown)


MattWestb:
The manufacture and board name token is normal terminated with one space.
But if writing one long string you is getting one warning (from Commander) that is no space between the token but its being written OK in the device flash.

Example for 15 charter long token that is not generating error (16 do but working OK in ZHA):
[MW-Marcus-RNP.txt](https://github.com/zigpy/bellows/files/9183609/MW-Marcus-RNP.txt)

puddly:
These tokens seem to be 16 bytes long. Maybe Commander adds `\x00` at the end to terminate the string, which means that you will have only 15 characters to work with.

Both should work with bellows but I think the `\x00` at the end is the "correct" way.

MattWestb:
I think the same that commander adding the `\x00` for strings that is under 16 chaster but i thing the flash is not having it then using 16 chaster tokens so dont make it braking if its not coming on stop charter 👍

Great work as normal from you Puddly !!!

Adminiuga (approved):
LGTM

anbugge:
Your assumptions are correct :)

This was actually a change in Commander a few years ago. From the change log:

* The MFG string tokens "MFG_STRING" and "MFG_BOARD_NAME" will now be zero terminated when written by Commander
    * Previously the string was unterminated, and unused token space is filled with 0xFF
    * If the string is exactly 16 bytes, filling the token, a warning will be raised and the string will be unterminated

PR #480, opened 2022-08-08
Probe with 115200 baud first

puddly:
All newer adapters are using 115200 baud, not 57600 (which I believe is only used by the old HUSBZB-1?). Changing the default to 115200 will break CLI tools and require a ZHA configuration migration so this is a slightly more gradual change.

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -54,7 +54,7 @@ class EZSP:
     @classmethod
     async def probe(cls, device_config: Dict) -> bool | dict[str, int | str | bool]:
         """Probe port for the device presence."""
-        for config in (device_config, {**device_config, CONF_DEVICE_BAUDRATE: 115200}):
+        for config in ({**device_config, CONF_DEVICE_BAUDRATE: 115200}, device_config):
             ezsp = cls(SCHEMA_DEVICE(config))
             try:
                 await asyncio.wait_for(ezsp._probe(), timeout=PROBE_TIMEOUT)


MattWestb:
Gary have cooked one firmware for HUSBZB-1 with 115200 SW and its looks working OK but its not recomeded of our HUSBZB specialist.
Its looks like all Silabs SOCs can using the higher baud rate without problems and i running RCP firmware with software flow control at 460800 and is stable :-)))

If the user is using recommended EZSP 6.7.x or newer we have 115200 SW firmware but no with 57600 and HW flow conrol (only EM53X).

Hedda:
> Its looks like all Silabs SOCs can using the higher baud rate without problems and i running RCP firmware with software flow control at 460800 and is stable :-)))
> 
> If the user is using recommended EZSP 6.7.x or newer we have 115200 SW firmware but no with 57600 and HW flow conrol (only EM53X).

That is correct. All Silicon Labs SoC hardware can technically support 115200 baud rate according to their data sheet specifications, but all older EM35x/EM35xx and ETRX35x (i.e. their "Ember" series of SoCs) basically all shipped with firmware that had 57600 baud rate as default as per old industry standard, while all newer EFR32MGxx ("Mighty Gecko" series of SoCs) basically all shipped with firmware that has 115200 baud rate as informally standardized by the DIY home automation community.

bellows README.md section that mention this -> https://github.com/zigpy/bellows/blob/dev/README.md#firmware-requirement

Examples of older adapters still commonly available and used which ship with 57600 baud rate as default from the factory:

- Nortek GoControl QuickStick Combo Model HUSBZB-1 Zigbee USB Adapter based on EM3581 (Ember 3581).
- Bitron Video AV2010/10, which is also sold as "Smabit BV AV2010/10 USB-Stick", "Telekom Magenta Stick", and "BitronVideo Funkstick") Zigbee USB stick based on Silicon Labs EM3587 (Ember 3587).
- Cortet ZM3588S-USB or Cortet ZM357S-USB EM35x USB Sticks by CEL based on Silabs EM3588 (Ember 3588) and EM357  (Ember 357)  "Ember" families of chips.

PR #485, opened 2022-08-20
Handle proactive startup reset frames

puddly:
If a coordinator itself sends a reset frame on startup and we have not yet communicated with the device, treat it as if we have reset the coordinator. This speeds up connection times to `zigbeed` considerably.

```python
# Event loop start
2022-08-20 11:23:18.740 root asyncio DEBUG Using selector: KqueueSelector

# Proactive reset received
2022-08-20 11:23:18.777 root bellows.uart DEBUG RSTACK Version: 2 Reason: RESET_SOFTWARE frame: b'c1020b0a527e'
2022-08-20 11:23:18.779 root bellows.ezsp DEBUG Received a reset on startup, not resetting again

# Connection works, starting up EZSP (0.040s later!)
2022-08-20 11:23:18.780 root bellows.ezsp.protocol DEBUG Send command version: (4,)
```

The few existing EFR32 network coordinators will experience a one second startup delay, since we can't distinguish a zigbeed coordinator from a typical ESP32 one.

@agners


MattWestb:
Its possible checking  EZSP is 7.0 or higher and Silabs addon is loaded.
But if running one RCP for thread and one NCP for Zigbee with EZSP 7.0 or higher is being false (im ruing one of this combination for test with Billy as RCP and Markus as NCP but with  EZSP 6.10.3).

Have not testing SkyConnect with RCP firmware but is coming (its running standard EZSP 6.10.3).

puddly:
> Its possible checking EZSP is 7.0 or higher and Silabs addon is loaded.

Do you mean there's a way to ask EZSP if it's running via RCP? That would be helpful.

MattWestb:
I think its not possible but if EZSP is 7.0 or higher it can being RCP but HA is knowing if Silabs addon is up and running and i think it shall being possible getting status is Zigbeed is up and running or if its not in the container then its one "normal" Linux container.
Or reading the silabs addon config that is have saved on the host system if its have the NVR then one Zigbeed is / was running.

PR #488, opened 2022-08-22
Use common EZSP init

agners:
This allows to use `bellows` CLI command with `zigbeed`.


agners:
Closing as this is handled by #487.

PR #498, opened 2022-10-07
Unset APS flags when sending broadcasts and multicasts

puddly:
Possibly fixes https://github.com/home-assistant/core/issues/79832

Also adds a tiny startup delay.


puddly:
This won't fix the problem, broadcasts just sometimes return a failure status code.

Issue #504, opened 2022-11-02
APS ACK ( retry) is always set


pipiche38:
In send_packet() APS_OPTION_RETRY is always set despite packet.tx_options setting.

https://github.com/zigpy/bellows/blob/a1c9d1ae1b62ffd955abd4b1ca37cedd43f8d59e/bellows/zigbee/application.py#L711

I'll try to submit a PR to make the aps_frame.options depending on packet.tx_options setting

pipiche38:
thanks for pushing the PR.
Let me know when you'll push to master

PR #505, opened 2022-11-02
make ACK dependents on packet.tx_options

pipiche38:
To prevent send_packet to always set APS_OPTION_RETRY check if packet tx_options requires or not ACK. Addressing #504


pipiche38:
Please let me know if there is any issue preventing merging that PR.
I'll correct it

puddly:
I will test this out later today

puddly:
I can confirm with Wireshark that this does indeed make bellows respect the zigpy ACK options. Thanks!

PR #506, opened 2022-11-06
Fix CI failures caused by #505

puddly:
For some reason CI did not run at all for this PR but it was still allowed to be merged.


MattWestb:
Black hacker fest without CI !!!!

PR #507, opened 2022-11-07
Allow disabling the bellows UART thread

puddly:
Not all environments support or need threading, namely https://pyscript.net/ and command line applications:

```yaml
zha:
  zigpy_config:
    use_thread: false
```

Once some hard-to-reproduce, low-level bugs are ironed out of the bellows asyncio thread, I think it would be beneficial to move it into zigpy.

### DIFF
--- a/bellows/config/__init__.py
+++ b/bellows/config/__init__.py
@@ -17,6 +17,7 @@ from zigpy.config import (  # noqa: F401 pylint: disable=unused-import
 )
 
 CONF_DEVICE_BAUDRATE = "baudrate"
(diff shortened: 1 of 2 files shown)


pipiche38:
What do you recommend ? I see that the default is to use the thread. In fact what is the benefit to use it ?

puddly:
> In fact what is the benefit to use it ?

There's no reason not to use it. If your event loop is busy or has stuttering, it will break EZSP communication with the radio. The only reason I'm adding this option is to allow bellows to be used in environments where threading is not implemented.

PR #510, opened 2022-11-10
Recalculate bad FCS for protocol versions 8 or newer

Adminiuga:
Recalculate FCS for EZSP version 8, 9 or newer when capturing ZigBee traffic.


puddly (approved):
Looks good to me. I can confirm this bug is still present with an EZSP 9 coordinator...

PR #520, opened 2023-01-06
Handle new Xiaomi IEEE prefix

puddly:
Might fix https://github.com/home-assistant/core/issues/85179


puddly:
Should we just add all of these? https://macaddress.io/statistics/company/14568

MattWestb:
I was not finding all prefixes for lumi that little strange.
We dont  knowing if they is using the IEEE for Zigbee / thread devices or BT, WiFi and Eth chips but if you is adding all it shall being future safe.
You can also taking on look wot deCONZ have putting in there code.

MattWestb:
DE is only using  2:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/674d17143b0467d8a9d14d6f876cee753aa445b1/zdp/zdp_handlers.cpp#L297-L300

puddly:
Interesting. Maybe `04:CF:FC` in bellows is a typo and should be `04:CF:8C` instead.

On the other hand, in the linked Home Assistant Core issue, if everything worked with the Conbee, it seems like its firmware might have a list as well?

TheJulianJES:
Yeah, at least with ZHA, Conbee sticks need to have their firmware updated most of the time, or they're not sending the Xiaomi codes.

Z-Stack also uses the same addresses which deCONZ has: https://github.com/Koenkk/Z-Stack-firmware/blob/3406c6a0cd2b755f302f4cacd07b37497a86cc08/coordinator/Z-Stack_3.x.0/firmware.patch#L551-L552

MattWestb:
So 6 case saying saying typo = more then likely.
Then the question shall wee adding many more IEEE ranges and making the system slower for all EZSP and deCONZ adapters then its not needed ?

MattWestb:
Have you checking the code in the radio libs for deCONZ and TI if its also is OK ?

puddly:
Bellows is the only library that handles these so I think we are fine. Z-Stack firmware is patched (https://github.com/zigpy/bellows/pull/520#issuecomment-1373927206) and the Conbee firmware seems to do it on its own.

PR #526, opened 2023-02-03
Handle startup reset exception being `None`

puddly:
https://github.com/zigpy/bellows/pull/524 fails to apply when the connection is gracefully closed by the server and `exc=None`.

Related: https://github.com/home-assistant/core/issues/86800


MattWestb:
Is it enough changing version in `/usr/src/homeassistant/homeassistant/components/zha/manifest.json` for getting ZHA using the updated 0.34.8 or do i need changing more places ??

puddly:
That should be enough. Make sure to restart HA and it will install the new version.

PR #463, opened 2022-06-13
Update README.md to mention EZSP v9 and EZSP v10

Hedda:
Update README.md to mention EZSP v9 Protocol Version as default in Silabs Zigbee EmberZNet 7.1.x firmware -> https://github.com/zigpy/bellows/pull/462

Update: Also updated this PR to in addition include mentioning of EZSP v10 as added in puddly's PR https://github.com/zigpy/bellows/pull/529

Silicon Labs Gecko SDK 4.1.0 (GSDK 4.1) was released a couple of days ago and with it Zigbee EmberZNet SDK 7.1.0.0 and tube0013 submitted a patch with initial support in PR https://github.com/zigpy/bellows/pull/462

https://github.com/SiliconLabs/gecko_sdk/releases

https://www.silabs.com/documents/public/release-notes/emberznet-release-notes-7.1.0.0.pdf

https://web.archive.org/web/20220401230237/https://www.silabs.com/documents/public/user-guides/ug100-ezsp-reference-guide.pdf

https://www.silabs.com/documents/public/user-guides/ug100-ezsp-reference-guide.pdf

EZSP protocol version has been incremented to EZSP v10 in Gecko SDK v4.2 / Zigbee EmberZNet v7.2 and EZSP v10 was not mentioned in the latest UG100 (EZSP Reference Guide) documentation from Silicon Labs which SiLabs only updated for EZSP v9 and state that no additions were added between EmberZNet PRO Release 7.1.1 and 7.2.0:

Gecko SDK v4.2:

https://github.com/SiliconLabs/gecko_sdk/blob/gsdk_4.2/protocol/zigbee/app/util/ezsp/ezsp-protocol.h

* EZSP_PROTOCOL_VERSION = 10

https://github.com/SiliconLabs/gecko_sdk/blob/gsdk_4.2/protocol/zigbee/app/util/ezsp/ezsp-protocol.h

https://www.silabs.com/documents/public/user-guides/ug100-ezsp-reference-guide.pdf

PR #529, opened 2023-02-17
EZSP v10

puddly:
The only major addition is `incomingNetworkStatusHandler`, which was previously `changeSourceRouteHandler`. I haven't seen any use for it:

```Python
DEBUG (MainThread) [bellows.ezsp.protocol] Application frame received incomingNetworkStatusHandler: [<EmberStackError.ROUTE_ERROR_MANY_TO_ONE_ROUTE_FAILURE: 12>, 0xafa8]
DEBUG (MainThread) [bellows.ezsp.protocol] Application frame received incomingNetworkStatusHandler: [<EmberStackError.ROUTE_ERROR_INDIRECT_TRANSACTION_EXPIRY: 6>, 0x9253]
```


bellows/ezsp/v10/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+from . import commands, config, types as v10_types
+from .. import protocol
+
+EZSP_VERSION = 10
+LOGGER = logging.getLogger(__name__)
+
+
+class EZSPv10(protocol.ProtocolHandler):
+    """EZSP Version 8 Protocol version handler."""
```
TheJulianJES, line 18:
```suggestion
class EZSPv10(protocol.ProtocolHandler):
    """EZSP Version 10 Protocol version handler."""
```

MattWestb:
In the current `UG100: EZSP Reference Guide` its the only change but its still V9 and is not updated.
I can being that V9 is not making the change and some is getting problems that the old command is used and not the new one and Silabs was bumping the protocol for blocking use of the old command that is making problems in some system (like Z2M).

Shall i posting one question in Silabs community and asking for information ??
If its not working its also possible putting one support ticket but i think it shall not being needed.

Hedda:
> The only major addition is `incomingNetworkStatusHandler`, which was previously `changeSourceRouteHandler`. I haven't seen any use for it

https://github.com/home-assistant/core/issues/85822 does mention those?

Hedda:
For reference also see https://github.com/Koenkk/zigbee-herdsman/pull/640 and https://github.com/Koenkk/zigbee2mqtt/issues/15724 and https://github.com/home-assistant/core/issues/85822

Hedda:
FYI, updated PR https://github.com/zigpy/bellows/pull/463 which adds a mentioning of EZSP v9 and EZSP v10 in the README.md for bellows.

MattWestb:
Time for one more update ;-)
`Protocol version 11 is not supported, using version 10 instead`
Its little to much coding for my so i letting it for some one that can doing it right from the geginning.

PS: Looks like the updated addon is using the save version of CPC V3 so GSDK 4.2.1.0 is still working and shall doing with dongle with 4.2.2.0 on addon 1.0.1

PR #537, opened 2023-03-19
Only wait for send status confirmation for unicasts

puddly:
This allows for the downstream application to send two consecutive group commands without waiting for a send confirmation for the first.

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -785,17 +785,18 @@ class ControllerApplication(zigpy.application.ControllerApplication):
                         status,
                     )
 
+                # Only throw a delivery exception for packets sent with NWK addressing.
+                # https://github.com/home-assistant/core/issues/79832
+                # Broadcasts/multicasts don't have ACKs or confirmations either.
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Changes seem to work as expected btw

pipiche38:
I don't know how it works with ezsp, but my experience on zigate is that by relying on this we were exhausting internal resources PDUs
So it might be interesting to see the impact on large heavy network.
And many having an easy way to disable this

puddly:
On EmberZNet the only effect is that you run into `NETWORK_BUSY` more quickly, which is expected since more group commands can be sent at once (e.g. group light color commands usually send `on()` + `move_to_color()`).

pipiche38:
And which layer handle the NETWORK_BUSY issue ? Is zigpy will do the retry or is that the above layer ?

puddly:
Bellows tries three times if it's a transient error: https://github.com/zigpy/bellows/blob/e700caf2a50454d0060c3fcd5f67f4b49ebfce46/bellows/zigbee/application.py#L760-L786

PR #550, opened 2023-04-18
Prefer firmware config defaults and only grow them when necessary

### LINKED ISSUE #543: Bellows overrides firmware settings with conservative values
Currently bellows overrides firmware settings for a lot of ezsp settings to very conservative settings. I think historically this was to combat poor firmware. But with newer devices based on the MG12/MG21/MG24 these higher fw settings are ignored unless configured in configuration.yaml.

This applies to the [default settings from the Nabu Casa built firmware](https://github.com/NabuCasa/silabs-firmware/wiki/Zigbee-EmberZNet-NCP-firmware-configuration) which I also use with TubesZB EFR32 devices. 

For example fw built supporting 200 source routes is set to a much lower value:

```
2023-04-02 10:19:51.067 DEBUG (MainThread) [bellows.ezsp.protocol] Application frame received setConfigurationValue: [<EzspStatus.SUCCESS: 0>]
2023-04-02 10:19:51.068 DEBUG (MainThread) [bellows.ezsp.protocol] Send command setConfigurationValue: (<EzspConfigId.CONFIG_SOURCE_ROUTE_TABLE_SIZE: 26>, 16)
```

These defaults handicap the ncp unless a user is technically proficient enough to configure all the setting in configuration.yaml:

image

How bellows handles defaults ezsp settings vs actual fw settings needs to be re-evaluated - also considering the change in 7.x (I think) that runtime configuration can only decrease settings not increase them.

puddly:
Default config is moved out of the protocol-specific definitions and into the main bellows application. With this change, two types of config are introduced:

 1. overrides: this config will always be set. For example, `CONFIG_TC_REJOINS_USING_WELL_KNOWN_KEY_TIMEOUT_S` will always be set to an exact value.
 2. minimum config: this config will only be changed if the current value is too low. Otherwise, the default value (set in firmware) is left alone.

Any of this behavior can be overridden by either overriding the config with an explicit value within your configuration, or by setting it to `None` to prevent bellows from changing it at all.

Fixes #543.


pipiche38:
Just checking, what does that means from a compatibility standpoint with the previous version ?

If I understood your PR, prior we had some restrictive values and now, by default we are following the default firmware values .

puddly:
> If I understood your PR, prior we had some restrictive values and now, by default we are following the default firmware values .

Yes. For example, if the firmware sets `CONFIG_MULTICAST_TABLE_SIZE` to `100`, then it will be left at `100`. If the firmware sets it to `2`, it will be grown to `16`. The old behavior is to always set it to `16`.

PR #564, opened 2023-06-12
Use a longer timeout when probing TCP coordinators

puddly:
Multi-PAN specifically takes too long and fails to `probe` most of the time. This is independent from `connect`, which works correctly.


MattWestb:
That i think its one good fix !!!!

PR #565, opened 2023-06-12
Write custom EUI64 to NV3 storage if available

puddly:
From the Gecko SDK 4.3.0 release notes:

> Zigbeed now includes an implementation of `emberGetRestoredEui64()` which loads the `CREATOR_STACK_RESTORED_EUI64` token
> from the `host_token.nvm` file.

This finally bypasses the dreaded `i_understand_i_can_update_eui64_only_once_and_i_still_want_to_do_it` limitation, allowing for a coordinator's IEEE address to be overwritten without any issues.

---

TODO:

- [x] Test with RCP firmware with `zigbeed`: works with the latest Gecko SDK release (v4.3.0).
- [x] Does it actually work? Is this IEEE address actually used in all tables and during network management?
- [x] Investigate if this works with NCP firmware (probably requires the NV3 component to be compiled into the firmware)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
           - --quiet-level=2
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.2.0
+    rev: v1.3.0
     hooks:
       - id: mypy
         additional_dependencies:
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -371,11 +371,83 @@ class EZSP:
             version = "unknown stack version"
         return tokens[0], tokens[1], version
 
-    async def can_write_custom_eui64(self) -> bool:
-        """Checks if the write-once custom EUI64 token has been written."""
-        (custom_eui_64,) = await self.getMfgToken(t.EzspMfgTokenId.MFG_CUSTOM_EUI_64)
+    async def _get_nv3_restored_eui64_key(self) -> t.NV3KeyId | None:
+        """Get the NV3 key for the device's restored EUI64, if one exists."""
+        try:
+            # If the EZSP version doesn't have `getTokenData`, it doesn't implement NV3
+            self.getTokenData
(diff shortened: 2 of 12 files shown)


MattWestb:
I think in GSDK 4.3.0.0 / EZSP 7.3.0.0 (or it was in the RCP part on the host side) its shall being possible loading it from software but i was not reading the dilates but it was very mush deep thing that only you can knowing how its working and if its possible implanting.
One thing is that they starting encrypting the key storage but i think we is reading it with commands that if doing it transparent.

Edit: I was not reading all and its looks you have implanting it !!!

Great work done as normal !!

agners:
> From the Gecko SDK 4.3.0 release notes:
> 
> > Zigbeed now includes an implementation of `emberGetRestoredEui64()` which loads the `CREATOR_STACK_RESTORED_EUI64` token
> > from the `host_token.nvm` file.

Isn't that referring to the Multiprotcool/Multi-PAN setup only (as in with `zigbeed`?

> 
> This finally bypasses the dreaded `i_understand_i_can_update_eui64_only_once_and_i_still_want_to_do_it` limitation, allowing for a coordinator's IEEE address to be overwritten without any issues.
> 

So how does that work/relate with regular NCP controllers using the EmberZNet stack running on the micro-controller?

puddly:
@agners The most recent commit implements it for NCP firmware. Corresponding PR for the firmware builder: https://github.com/NabuCasa/silabs-firmware-builder/pull/22

I've yet to actually test these in a real network beyond ensuring the IEEE persists through backup and restore.

puddly:
I migrated my home network to an old SkyConnect that has already had its EUI64 replaced with an incorrect value. Everything works fine so far with the NV3 token storage after flashing 7.3.0.0 NCP with the extra component.

dmulcahey:
Nice!!

MattWestb:
@puddly If you is having one SWD probe you can writing one empty file to the "userdata" or erasing  the flash pages with command and the token of the burned is going away and can being burned agen.

PR #567, opened 2023-06-22
EZSP v12

puddly:
There are a few new commands related to key handling and green power.


MattWestb:
Only  "some" and in the end of the summer V13 is coming with more work for our **Puddly** !!
Good that Bellow and EZSP always is backward compatible so still working with new version as long the format is being changed.

### Great work done !!

PR #571, opened 2023-06-30
Handle `getTokenData` command missing

puddly:
Some versions of EmberZNet do not include the command at all, even if it is defined in the EZSP version.

https://github.com/home-assistant/core/issues/95632

### DIFF
--- a/bellows/exception.py
+++ b/bellows/exception.py
@@ -5,5 +5,9 @@ class EzspError(APIException):
     pass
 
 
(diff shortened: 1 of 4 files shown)


puddly:
Tested with HUSBZB-1, SkyConnect running EZSP v7.1.3.0 and v7.3.3.0, and RCP v7.3.3.0.

balloob (approved):

PR #576, opened 2023-08-04
Force TX after CCA failure on EZSP v7+

puddly:
`EMBER_VALUE_FORCE_TX_AFTER_FAILED_CCA_ATTEMPTS` seems to work in the latest Gecko SDK release! This allows the radio to force a transmit, even after CCA failures.

Normally this sounds like a bad idea but the alternative is to allow the radio to silently fail with a generic "delivery error" (TODO: figure out a way to infer the actual error code).

### DIFF
--- a/bellows/ezsp/config.py
+++ b/bellows/ezsp/config.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import dataclasses
+import typing
 
 import bellows.ezsp.v4.types as types_v4
 import bellows.ezsp.v6.types as types_v6
+import bellows.ezsp.v7.types as types_v7
 import bellows.types as t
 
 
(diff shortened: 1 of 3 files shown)


puddly:
It appears that this change is compatible with many more versions of EmberZNet than I thought: reading the value with `getValue` fails with `EzspStatus.ERROR_INVALID_ID`, but the write will still succeed. I've confirmed that the CCA bypassing works as far back as EmberZNet 7.1.1.0, and I'm sure previous releases too, meaning this PR will affect most modern EZSP adapters.

puddly:
And with the adapter in a noisy environment, here are some traffic captures:

**With CCA bypass** (coordinator responds to beacon request)

image

---

**Without CCA bypass** (coordinator doesn't respond)

image

---

A device was only able to join (successfully might I add) in the first capture.

MattWestb:
Great finding and work done our master HACKER !!!!

PR #578, opened 2023-08-30
Fix startup broken by #577

puddly:
#577 reorganized startup to allow for a simpler runtime reset (required for rebooting after writing config) but unfortunately omitted calls to `register_endpoints` and `write_config`, severely breaking it. This PR should address this.

CC @TheJulianJES


TheJulianJES:
Yep, seems to have fixed the issue 🎉

PR #579, opened 2023-09-01
Reset after NVRAM changes

puddly:
Bellows can sometimes have issues forming new networks: settings are successfully written, the network starts up, a backup afterwards works, but upon reset the settings are lost/revert.

It seems that we need to write our EmberZNet config after reset, in addition to resetting after every NVRAM change.

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -565,6 +565,16 @@ class EZSP:
                 ],
             }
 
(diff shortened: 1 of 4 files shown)


puddly:
Everything works with all of the coordinators I tested (including an ancient HUSBZB-1) but I'd appreciate a second check. I encountered this issue when working on https://github.com/home-assistant/core/pull/99482.

PR #589, opened 2023-10-23
Always send `version(4)` when resetting

puddly:
This is the root cause of https://github.com/home-assistant/core/issues/102503.

Firmware crashes were seen on EZSPv6 but not on the v8 stick I was testing when I initially added this special case. A HUSBZB-1 running EZSPv8, however, still crashes. This PR removes the protocol version limitation for removing this config.

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -183,25 +183,33 @@ class EZSP:
         LOGGER.debug("Resetting EZSP")
         self.stop_ezsp()
         await self._gw.reset()
+
+        # Always switch back to protocol v4 after a reset
+        self._switch_protocol_version(v4.EZSPv4.VERSION)
         self.start_ezsp()
+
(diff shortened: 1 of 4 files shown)


MattWestb:
I think EM35X and ÉFR32GM1X devices is getting short of RAM then stretching them to much but its also depends of the config of the firmware and its little different version around here.
MG12, MG13 and all MG2X shall not getting problems with RAM so fast then they is having plenty of it as long the firmware is OK configured. 
I only running one EZSP / NCP with MG1P (Billy) and its restarting little now and then but no big problems (all test system is running RCP).

puddly:
I'm renaming this PR because the problem is not what I initially suspected.

The *root* cause is...sending `version(4)` is *required* when you reset! Even though `version(8)` appears to work and the protocol version is upgraded, random crashes appear during startup. Sending `version(4)` before `version(8)` fixes the `CONFIG_PACKET_BUFFER_COUNT` issue as well.

PR #594, opened 2023-11-14
Correct gpepIncomingMessageHandler data signature

### LINKED ISSUE #593: EZSP docs incorrectly describe gpepIncomingMessageHandler data
Following the documentation unfortunately returns bad data for gpepIncomingMessageHandler. Fortunately, it looks like zigbee-herdsman gets it right [in their implementation](https://github.com/Koenkk/zigbee-herdsman/blob/09a24e8182e5b1ba56e81f32e8a2defb3d5d247f/src/adapter/ezsp/driver/commands.ts#L2205). I've dissected it at the data level and their implementation looks good.

konistehrad:
This assumes that it's the same across all versions. Of course, that's very difficult to validate given the state of EZSP documentation. However, on v8 this is absolutely correct, and I don't see a reason why it might be only v8 as opposed to everywhere else. Will close #593.

### DIFF
--- a/bellows/ezsp/v10/commands.py
+++ b/bellows/ezsp/v10/commands.py
@@ -628,19 +628,22 @@ COMMANDS = {
         0x00C5,
         (),
         (
-            t.EmberStatus,
-            t.uint8_t,
-            t.uint8_t,
-            t.EmberGpAddress,
+            t.EmberStatus,           # status
+            t.uint8_t,               # gpd link
+            t.uint8_t,               # sequence number
+            t.uint8_t,               # addrType
+            t.uint32_t,              # address
+            t.uint32_t,              # src ID
+            t.uint8_t,               # endpoint
             t.EmberGpSecurityLevel,
             t.EmberGpKeyType,
-            t.Bool,
-            t.Bool,
-            t.uint32_t,
-            t.uint8_t,
-            t.uint32_t,
-            t.EmberGpSinkListEntry,
-            t.LVBytes,
+            t.Bool,                  # auto commissioning
+            t.Bool,                  # rx capable?
+            t.uint32_t,              # security frame counter
+            t.uint8_t,               # gpd command id
+            t.uint32_t,              # MIC
+            t.uint8_t,               # proxy table index
+            t.LVBytes,               # optional payload
         ),
     ),
     "gpProxyTableGetEntry": (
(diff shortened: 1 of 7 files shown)


konistehrad:
Note that the structure was lifted wholesale from [zigbee-herdsman](https://github.com/Koenkk/zigbee-herdsman/blob/09a24e8182e5b1ba56e81f32e8a2defb3d5d247f/src/adapter/ezsp/driver/commands.ts#L2205) though I was able to verify their implement was correct at the data level and also polish it up a bit for us.

puddly:
Thanks! Here's what is in the SDK source:

```C
// A callback invoked by the ZigBee GP stack when a GPDF is received.
void ezspGpepIncomingMessageHandler(
  // The status of the GPDF receive.
  EmberStatus status,
  // The gpdLink value of the received GPDF.
  uint8_t gpdLink,
  // The GPDF sequence number.
  uint8_t sequenceNumber,
  // The address of the source GPD.
  EmberGpAddress *addr,
  // The security level of the received GPDF.
  EmberGpSecurityLevel gpdfSecurityLevel,
  // The securityKeyType used to decrypt/authenticate the incoming GPDF.
  EmberGpKeyType gpdfSecurityKeyType,
  // Whether the incoming GPDF had the auto-commissioning bit set.
  bool autoCommissioning,
  // Bidirectional information represented in bitfields, where bit0 holds
  // the rxAfterTx of incoming gpdf and bit1 holds if tx queue is available
  // for outgoing gpdf.
  uint8_t bidirectionalInfo,
  // The security frame counter of the incoming GDPF.
  uint32_t gpdSecurityFrameCounter,
  // The gpdCommandId of the incoming GPDF.
  uint8_t gpdCommandId,
  // The received MIC of the GPDF.
  uint32_t mic,
  // The proxy table index of the corresponding proxy table entry to the
  // incoming GPDF.
  uint8_t proxyTableIndex,
  // The length of the GPD command payload.
  uint8_t gpdCommandPayloadLength,
  // The GPD command payload.
  uint8_t *gpdCommandPayload)
```

konistehrad:
Actually I’m really sorry I pressed my EFR32 into running my home net while I wait for a replacement to ship. (EZSP does bespoke GPF processing that’s out of scope for my work right now.) I’ll dig back into this ASAP but it’ll be a bit.

PR #598, opened 2023-12-04
Remove UART thread

puddly:
Zigpy uses [pyserial-asyncio-fast](https://github.com/home-assistant-libs/pyserial-asyncio-fast) if it is installed. This fork of pyserial-asyncio fixes a bug where writes are buffered unnecessarily, essentially introducing unnecessary delays to serial writes (@pipiche38 you may want to add this as a dependency to benefit from it).

The purpose of the UART thread was to allow bellows to instantly respond to packets with an ACK and avoid timing problems caused by a busy main asyncio event loop. This problem is fixed by pyserial-asyncio-fast, as `protocol.data_received` synchronously parses any incoming data and immediately calls `transport.write`, without invoking the event loop at all.

In my local tests, 4608/4608 `transport.write` calls immediately resulted in a successful `write` syscall, meaning the protocol's write buffer was never filled at all.


pipiche38:
@puddly thanks for the ping. I had a look to the PR and you never refer to the pyserial-asyncio-fast. At what level is pyserial-asyncio-fast replacing pyserial-asyncio ?

puddly:
@pipiche38 Zigpy does it automatically if the package is available: https://github.com/zigpy/zigpy/blob/fb19ae5904d79bee43a0188636088499cc674a7b/zigpy/serial.py#L15-L20

PR #599, opened 2023-12-10
Reuse ZDO Initializers to create Endpoint objects on EZSP device

konistehrad:
We need access to a real GP `Cluster` type for parsing the incoming frames, but EZSP doesn't support polling the device for current endpoints. Instead, reuse the old ZDO initializers and assume the device state matches the requested ZDO descriptors.


konistehrad:
Not sure what to do about codecov here; the endpoint initializer is async but the application fixture looks to be synchronous and everywhere. Happy to fix it, but I'll need a little handholding.

konistehrad:
I used this patch on the green power test net for a couple days and everything seems 100%. Thanks again for the cleanup job.

konistehrad:
You love to see it. Thanks again!

PR #603, opened 2023-12-22
EZSP v13

### LINKED ISSUE #601: EZSP v13 (EZSP PROTOCOL VERSION 13) support in bellows?
EZSP protocol version been incremented to EZSP v13 in Silicon Labs Gecko SDK v4.4 (4.4.0.0) / Zigbee EmberZNet v7.4 (7.4.0.0):

https://github.com/SiliconLabs/gecko_sdk/releases/tag/v4.4.0

https://www.silabs.com/documents/public/release-notes/emberznet-release-notes-7.4.0.0.pdf

https://www.silabs.com/wireless/zigbee?tab=software#software

0x0D hexadecimal = 13 in decimal.

https://github.com/SiliconLabs/gecko_sdk/blob/gsdk_4.4/protocol/zigbee/app/util/ezsp/ezsp-protocol.h#L33

Not sure if any relevant changes or not?

puddly:
This release of EmberZNet drops support for reading/writing network keys using our existing methods, breaking compatibility. The new commands thankfully maintain feature parity and include new features, allowing us to fully restore APS link keys. This should make backup/restore 100% state-preserving for EmberZNet.


bellows/zigbee/application.py
```diff
@@ -421,35 +511,46 @@ async def write_network_info(
...
+
+            for child_eui64 in self.state.network_info.children:
+                if child_eui64 not in self.state.network_info.nwk_addresses:
+                    continue
+
+                status = await ezsp.setChildData(
+                    index,
+                    ezsp.types.ChildData(
+                        eui64=child_eui64,
+                        # XXX: why does EmberZNet also have `MOBILE_END_DEVICE`?
```
puddly: The SDK also defines `MOBILE_END_DEVICE` but I can't figure out what this refers to.

@MattWestb have you heard of this before?

puddly: Looks like a proprietary extension.

> Mobile end devices (EmberNodeType EMBER_MOBILE_END_DEVICE) is a sleepy end device with enhanced capabilities that enable it to change its physical location and quickly switch to a new parent router. This device type is an Ember modification to the basic ZigBee sleepy end device to provide added capabilities and management of mobile devices.


> Mobile End Devices - For devices that change parents (or move out of parent range) often, they can join as EMBER_MOBILE_END_DEVICE (selectable in AppBuilder) and our stack will store the child entries in RAM with a more aggressive timeout (EMBER_MOBILE_NODE_POLL_TIMEOUT) rather than writing to SimEEPROM memory each time the child comes and goes.

I think using `SLEEPY_END_DEVICE` is better, as in zigpy we don't keep track of the child type.

dmulcahey:
```
There are three types of sensor nodes: line powered sensors, sleepy
//  sensors, and mobile sensors. Line powered sensors route and are always
//  awake. They can also serve as parents to sleepy or mobile sensors.
//  Sleepy sensors sleep whenever they can to conserve power but they do
//  not route and rely on their parent when communicating with the network.
//  Mobile sensors are like sleepy sensors but are meant to move. mobile
//  sensors will wait a smaller amount of time (5 seconds) to switch parents
//  than sleepy sensors will (60 seconds) if their parent does not respond to
//  their polls.
```
I found this in a header for one of their samples

puddly:
This may also be an interesting command to explore. It's available in EZSP v9+:

image


```C
/** @brief Set the routing shortcut threshold to directly use a neighbor instead of
 * performing routing.
 *
 * @param costThresh the routing shortcut value that should be used to directly use a neighbor
 * instead of performing routing.
 *
 * @return ::return EMBER_BAD_ARGUMENT if the costThresh is not 1, 3, 5, 7
 * Returns ::EMBER_SUCCESS otherwise.
 */
EmberStatus emberSetRoutingShortcutThreshold(uint8_t costThresh);
```

> Values range from 1 for a good link to 7 for a bad link.

PR #611, opened 2024-02-09
Support firmware extensions

puddly:
Firmware side: https://github.com/NabuCasa/silabs-firmware-builder/pull/57

Support firmware extensions that expose functionality not present in stock EmberZNet. As a demo, I've implemented a custom EmberZNet packet filter that should ensure all group traffic is passed through without needing to add the NCP to the group beforehand. This removes the need for `bellows.multicast`.


MattWestb:
This is VERY interesting !!
If its working well we need patching all current production firmware that is possible but i think it worth it.
I have good contact with Garry R (down under) that is the maker or most free ones only Sonoff signed can being problems but the dev still having his git but have moving to other company.
Any link to your added code for SS 5 as extension ?

MattWestb:
By the way, filtering multicast frames commands that have the same TSN from the same source and destination shall it being in the firmware or in bellows then the factory one is not doing it so normally getting multiple events with the same command ?

puddly:
Firmware side is here: <https://github.com/NabuCasa/silabs-firmware-builder/pull/57>

You can try pre-built firmwares for the SkyConnect and Yellow as well: <https://github.com/NabuCasa/silabs-firmware-builder/actions/runs/8945075022#artifacts>

PR #612, opened 2024-02-12
Don't remove relays when there are route errors

dmulcahey:


dmulcahey:
@puddly codecov is drunk on this one

puddly (approved):

PR #616, opened 2024-04-12
Fix Python 3.12.3 regression with enums

puddly:
Python 3.12.3 recently released and (again) changed the behavior of the `enum` module: https://github.com/python/cpython/issues/116040

Zigpy isn't affected so I've replaced all bellows types with zigpy types to hopefully avoid these issues in the future.


puddly:
Looks like there was a pretty big bug with the old types:

```python
(Pdb) t.uint16_t.deserialize(b'\x06')
(6, b'')
```

This should not have deserialized.

puddly:
It looks like the integer parsing problem is pretty widespread. This branch runs for me on EZSP v13 but it needs to be runtime tested on all previous protocols too...

puddly:
Tested with protocol versions 4, 8, and 13. I think this is fine to merge.

PR #626, opened 2024-05-28
Source routing fixes

puddly:
Two fixes:

1. `setSourceRoute` is a no-op in EZSPv9 and above, the command does nothing. There is currently no known way to send packets with a specific source route in these EZSP versions. Maybe `emberOverrideAppendSourceRouteHandler` can help if we create a custom XNCP command to maintain a temporary table of source routes per NWK address?
2. According to Silicon Labs, IEEE address discovery broadcasts are their solution to the missing source route problem. The coordinator will use them in place of normal routing commands to discover routes to a specific device if the NCP isn't aware of a source route to it. Enabling normal route discovery will actually cause issues with routers, as they will flip between source routing and normal routing when receiving these packets.

CC @TheJulianJES


TheJulianJES:
Applied this fix on my production setup. Unfortunately, I'm still seeing "missing routes" when testing 7 minutes after rebooting HA (just unable to toggle certain lights). I'll do more testing later.

dmulcahey:
I have a branch somewhere where I was trying to manually send a MTOR request right after startup. Wonder if that is worth exploring more.

dmulcahey:
> I have a branch somewhere where I was trying to manually send a MTOR request right after startup. Wonder if that is worth exploring more.

https://github.com/zigpy/bellows/compare/dev...dmulcahey:bellows:dm/refresh-routes

https://github.com/zigpy/bellows/compare/dev...dmulcahey:bellows:dm/improve-source-routing

looks like I had 2 branches out there.... maybe some combo of this + tweaking will help. I never got too far w/ testing.

puddly:
I've implemented "dumb" source routing as a firmware extension. Alternate branches:

- https://github.com/NabuCasa/silabs-firmware-builder/pull/57 (you can download firmwares for the SkyConnect and Yellow once it finishes: https://github.com/NabuCasa/silabs-firmware-builder/actions/runs/9275212477?pr=57)
- https://github.com/zigpy/bellows/pull/611

Network startup and mass command sending is now fast 😄

PR #628, opened 2024-06-13
Prevent task cancellation from propagating to ASH

puddly:
Root cause of https://github.com/home-assistant/core/issues/119424.

During device joining, zigpy cancels a scheduled initialization task if the device re-joins during initialization (this is pretty common). Unfortunately, this cancellation propagates *all* the way down to the ASH sending task, causing the TX sequence number to increment without waiting for an acknowledgement. There is currently a firmware bug with EmberZNet and while ASH can support multiple pending frames at a time, in reality the stack crashes if the number is greater than one 😄.

This fix prevents an ASH send from being cancelled by using `asyncio.shield`, which schedules it in a task. Incrementing the TX sequence number after a frame has been sent will have similar issues because our last send may not have been ACKed. An alternative to this approach would be to avoid using coroutines entirely for sending and make the ASH protocol implementation rely on event loop callbacks for timeouts.

### DIFF
--- a/bellows/ash.py
+++ b/bellows/ash.py
@@ -57,7 +57,7 @@ T_REMOTE_NOTRDY = 1.0
 
 # Maximum number of DATA frames the NCP can transmit without having received
 # acknowledgements
-TX_K = 1
+TX_K = 1  # TODO: investigate why this cannot be raised without causing a firmware crash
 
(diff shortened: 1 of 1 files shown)


tube0013:
2 hours in this is working well, no more NCP failures.

TheJulianJES (approved):

PR #630, opened 2024-06-13
Fix blocking metadata version call

TheJulianJES:
This moves the blocking metadata version call outside of the `load_network_info` coroutine that's run at startup. 
I'll do this for the other radio libraries too, just want to confirm if we want to have that constant named `LIB_VERSION`, just `VERSION`, `PACKAGE_VERSION`, or `BELLOWS_VERSION`? 

The test still blocks when checking it, but that should be fine.

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -65,6 +65,7 @@ IEEE_PREFIX_MFG_ID = {
     "54:EF:44": 0x115F,  # Lumi
 }
 
+LIB_VERSION = importlib.metadata.version("bellows")
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


puddly (approved):

puddly:
I think I've only ever seen `top_level_module.__version__`. Putting it in `module.application.zigbee.LIB_VERSION` makes it pretty much library-internal.

TheJulianJES:
> I think I've only ever seen `top_level_module.__version__`. Putting it in `module.application.zigbee.LIB_VERSION` makes it pretty much library-internal.

Yeah, I don't expect anyone else to use this. This is just to fix the blocking call.
I just used `LIB_VERSION` as the name, so it's clear that we don't access the EZSP version or whatever.
Or do you want something else?

puddly:
Nope, it's all good the way it is. Thanks!

PR #631, opened 2024-06-19
EZSP v14

puddly:
This is going to be a huge change. Very WIP, only backup and restore is working.

- The command format for sending packets seems to have changed significantly.
- Parameters for most EZSP commands have been reordered.
- The error codes returned by the firmware have also completely changed and are not compatible with previous versions.
- Nothing is documented 😄  (https://github.com/SiliconLabs/simplicity_sdk/issues/11)


puddly:
This PR is in a state to be able to run successfully on my home network.

However, EmberZNet 8.0.0.0 is completely broken:

1. LQI and RSSI just do not contain valid values for any packet.
2. Unicast performance is abysmal and most sends do not succeed once more than a few packets have been sent at once.

This PR needs to be merged so that `universal-silabs-flasher` can enter the bootloader to downgrade from 8.0.0.0 but I'm otherwise not going to investigate EmberZNet 8.0.0.0 further.

puddly:
Project coverage shrinks just a little bit with this PR but unfortunately there's no sensible way to actually deal with EZSP v14 command processing without rewriting all EZSP commands to use named parameters.

Hedda:
Also see comments by Zigbee2MQTT/zigbee-herdsman ember developer Nerivec in https://github.com/SiliconLabs/simplicity_sdk/issues/13

PR #633, opened 2024-07-14
Cleanly shut down the serial port on disconnect 

puddly:
Sample implementation of https://github.com/zigpy/zigpy/pull/1425


ryanwinter:
I'm thinking to Dutch my husbzb and getting a sonoff v3 instead because I suspect that this change has just surfaced an underlying issue with my controller in that it is randomly resetting.

PR #635, opened 2024-07-21
Reduce command duplication

puddly:
This PR finishes the "derives from the previous version" format that we've been using for EZSP v8-v14.


puddly:
I believe that https://github.com/zigpy/bellows/pull/635/commits/18d1d524830c536666e22a2a403a56718a91e003 is going to be a safe change: I've not seen a single SDK version where the sent command parameter is actually 16 bits. The SDK uses 16-bit integer constants but the command is always serialized and deserialized to/from an 8-bit integer. I think the fact that EZSP is lenient with command payload lengths and that we use little endian encoding has sort of "hidden" this problem:

```python
In [1]: (0x0012).to_bytes(2, 'little')
Out[1]: b'\x12\x00'

In [2]: (0x0012).to_bytes(1, 'little')
Out[2]: b'\x12'
```

PR #637, opened 2024-07-23
Fix bellows CLI

puddly:
I've removed network backup and restore from the bellows CLI. It is too complex to maintain and should be kept in zigpy-cli.

The rest of the tools were spot checked and run.


puddly:
In the future, we need to remove the bellows CLI entirely. This has a few blockers:

- Add the `--i-understand-...` flag to zigpy-cli, maybe call it `--burn`?
- Implement radio-agnostic sniffing.
- Add more options for network formation (forcing a PAN ID, EPID, channel, key, IEEE address, and so on).
- Maybe add a way to send ZCL commands to devices?

PR #646, opened 2024-08-12
Improve transmission reliability for sleepy end-devices

puddly:
End devices are flagged as needing an "extended timeout" by their IEEE but we often times never inform the stack of the NWK -> IEEE mapping. We fix this by injecting a random entry into the address table if one does not already exist.

### DIFF
--- a/bellows/ezsp/v4/__init__.py
+++ b/bellows/ezsp/v4/__init__.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import logging
+import random
 from typing import AsyncGenerator, Iterable
 
 import voluptuous as vol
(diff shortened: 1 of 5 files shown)


bellows/ezsp/v4/__init__.py
```diff
@@ -193,3 +194,41 @@ async def read_counters(self) -> dict[t.EmberCounterType, t.uint16_t]:
...
+        (node_id,) = await self.lookupNodeIdByEui64(eui64=ieee)
+
+        # Check to see if we have an address table entry
+        if node_id != 0xFFFF:
+            await self.setExtendedTimeout(
+                remoteEui64=ieee, extendedTimeout=extended_timeout
+            )
+            return
+
+        (status, addr_table_size) = await self.getConfigurationValue(
```
dmulcahey: Can we capture this when applying the config so it is only done once when the application starts?

dmulcahey (approved):

PR #649, opened 2024-09-09
Move EZSP send lock from `EZSP` to individual protocol handlers

puddly:
We recently moved a lot of code into individual EZSP protocol handlers. This introduced a "hole" where protocol handlers could run multiple commands at once within a single semaphore.

This PR moves the semaphore into the protocol handlers, closing this hole. I've also adjusted protocol timeouts to better handle TCP failures, making communication with TCP coordinators a bit more resilient. I think after this PR we may be able to immediately disconnect on watchdog failure as opposed to allowing up to five to fail.

CC @tube0013

### DIFF
--- a/bellows/ezsp/protocol.py
+++ b/bellows/ezsp/protocol.py
@@ -90,27 +90,28 @@ class ProtocolHandler(abc.ABC):
 
     async def command(self, name, *args, **kwargs) -> Any:
         """Serialize command and send it."""
-        delay_time = 0
-        was_delayed = False
+        delayed = False
+        send_time = None
 
         if self._send_semaphore.locked():
+            delayed = True
+            send_time = time.monotonic()
+
             LOGGER.debug(
                 "Send semaphore is locked, delaying before sending %s(%r, %r)",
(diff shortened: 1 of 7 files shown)


bellows/ezsp/protocol.py
```diff
@@ -65,18 +73,58 @@ def _ezsp_frame_rx(self, data: bytes) -> tuple[int, int, bytes]:
...
-        data = self._ezsp_frame(name, *args, **kwargs)
-        cmd_id, _, rx_schema = self.COMMANDS[name]
-        future = asyncio.get_running_loop().create_future()
-        self._awaiting[self._seq] = (cmd_id, rx_schema, future)
-        self._seq = (self._seq + 1) % 256
-
-        async with asyncio_timeout(EZSP_CMD_TIMEOUT):
-            await self._gw.send_data(data)
-            return await future
+        delay_time = 0
```
dmulcahey: Should we call this attempted_time? Or something else that makes it clearer that it is the original time and not some sort of delay period?

PR #669, opened 2025-02-26
Add support for receiving basic fragmented messages

ThomasKulin:
I am not much of a developer but I have found your project to be very useful and would like to contribute a change that I made to implement fragmented message support. This likely needs to be expanded to respect the EzspConfigId.CONFIG_FRAGMENT_WINDOW_SIZE and EzspConfigId.CONFIG_FRAGMENT_DELAY_MS configuration parameters.

### DIFF
--- a/bellows/ezsp/fragmentation.py
+++ b/bellows/ezsp/fragmentation.py
@@ -1,0 +1,98 @@
+"""
+Implements APS fragmentation reassembly on the EZSP Host side,
+mirroring the logic from fragmentation.c in the EmberZNet stack.
+"""
+
+import asyncio
+import logging
+from collections import defaultdict
+from typing import Optional, Dict, Tuple
+
+LOGGER = logging.getLogger(__name__)
+
+# The maximum time (in seconds) we wait for all fragments of a given message.
+# If not all fragments arrive within this time, we discard the partial data.
+FRAGMENT_TIMEOUT = 10
+
+# store partial data keyed by (sender, aps_sequence, profile_id, cluster_id)
+FragmentKey = Tuple[int, int, int, int]
+
+class _FragmentEntry:
+    def __init__(self, fragment_count: int):
+        self.fragment_count = fragment_count
+        self.fragments_received = 0
+        self.fragment_data = {}
+        self.start_time = asyncio.get_event_loop().time()
+    
+    def add_fragment(self, index: int, data: bytes) -> None:
+        if index not in self.fragment_data:
+            self.fragment_data[index] = data
+            self.fragments_received += 1
+
+    def is_complete(self) -> bool:
+        return self.fragments_received == self.fragment_count
+
+    def assemble(self) -> bytes:
+        return b''.join(self.fragment_data[i] for i in sorted(self.fragment_data.keys()))
+
+class FragmentManager:
+    def __init__(self):
+        self._partial: Dict[FragmentKey, _FragmentEntry] = {}
+    
+    def handle_incoming_fragment(self, sender_nwk: int, aps_sequence: int, profile_id: int, cluster_id: int,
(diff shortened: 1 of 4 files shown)


puddly:
Interesting! Are fragmented frames something you're seeing with a consumer device, or is this a device that you're working on?

ThomasKulin:
I haven't seen them used in any consumer device so far. This is for something relating to solar inverters where Zigbee is used for data collection. Some of the data packets are large enough to necessitate fragmentation. This code works for my specific application where the device fragment window size is set to 1 (requiring an ack for each transmitted fragment). I have no way of testing with other window sizes. Feel free to recommend changes and I'll gladly make them!

bellows/ezsp/fragmentation.py
```diff
@@ -0,0 +1,98 @@
...
+        return self.fragments_received == self.fragment_count
+
+    def assemble(self) -> bytes:
+        return b''.join(self.fragment_data[i] for i in sorted(self.fragment_data.keys()))
+
+class FragmentManager:
+    def __init__(self):
+        self._partial: Dict[FragmentKey, _FragmentEntry] = {}
+    
+    def handle_incoming_fragment(self, sender_nwk: int, aps_sequence: int, profile_id: int, cluster_id: int,
```
puddly: For clarity, I think we should pre-split the `group_id` field into `fragment_count` and `fragment_index` parameters, just so this method isn't stuck being responsible for low-level parsing logic.

bellows/ezsp/fragmentation.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if entry.is_complete():
+            reassembled = entry.assemble()
+            del self._partial[key]
+            LOGGER.debug("Message reassembly complete. Total length=%d", len(reassembled))
+            return (True, reassembled, fragment_count, fragment_index)
+        else:
+            return (False, None, fragment_count, fragment_index)
+
+    def cleanup_expired(self) -> None:
```
puddly: A slightly different pattern that I think might eliminate the need for the perpetually running 5s loop would be to create and cancel asyncio timers on the fly during reception. Something like this:

```python
class FragmentManager:
    def __init__(self):
        self._partial: Dict[FragmentKey, _FragmentEntry] = {}
        self._cleanup_timers: Dict[FragmentKey, asyncio.TimerHandle] = {}

    def handle_incoming_fragment(...):
        ...
        # Cancel the current timer, if there is one
        if key in self._cleanup_timers:
            self._cleanup_timers[key].cancel()
            del self._cleanup_timers[key]

        # This would allow messages to be reconstructed as long as *some* fragment is
        # received within `FRAGMENT_TIMEOUT`, so it wouldn't be a global timeout.
        loop = asyncio.get_running_loop()
        self._cleanup_timers[key] = loop.call_soon(FRAGMENT_TIMEOUT, self.cleanup_fragment, key)
        ...

    def cleanup_fragment(self, key: FragmentKey) -> None:
        del self._partial[key]
        del self._cleanup_timers[key]
        LOGGER.debug("Removed stale fragment reassembly for key=%s", k)
```

puddly (changes requested):
I left a few comments regarding the cleanup code. Let me know if you have any questions.

Thanks again!

puddly:
For the failing CI, you can set up pre-commit to auto-format your code and point out issues that CI would detect:

```bash
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

This will also run every time you make a new Git commit, for future changes.

bellows/ezsp/fragmentation.py
ThomasKulin: Good idea. Done!

bellows/ezsp/fragmentation.py
ThomasKulin: This is much cleaner, thanks

PR #677, opened 2025-07-01
Autoprobe with 460800 baud

puddly:
I'm going to be using 460800 baud for newer adapters.

57600 is used by I think one or two sticks, I wonder if it's time to drop it? Right now autoprobing is quite slow due to the sheer number of combinations.

### DIFF
--- a/bellows/zigbee/application.py
+++ b/bellows/zigbee/application.py
@@ -80,6 +80,7 @@ class ControllerApplication(zigpy.application.ControllerApplication):
     _watchdog_period = 10
     _probe_configs = [
         {zigpy.config.CONF_DEVICE_BAUDRATE: 115200},
+        {zigpy.config.CONF_DEVICE_BAUDRATE: 460800},
         {zigpy.config.CONF_DEVICE_BAUDRATE: 57600},
     ]
 


MattWestb:
If i remember was old EM3X with original firmware using 57600 but the latest was compiled with 115200 but i think most of the EM3X dongles is now out then no current EZSP can being compiled for them plus need one tool-change that is very expensive and i think only Garry can do it in the community.
See old Wiki for hardware and EZSP version (all older then 6.7.8.0 is obselite) https://github.com/zigpy/zigpy/wiki/Coordinator-Firmware-Updates#recommended-firmware-versions-for-ezsp-coordinators 

PS: most little older hardware with updated EZSP is using SW flow control at 115200 and only some hardware flow control so still importing have the 115200 probing.

puddly:
The only stick that I have is the HUSBZB-1 with the original firmware (57600). All of the updated firmwares I think switch it to 115200?

pipiche38:
Would it make sense to start at 460800  then if no communication established switch to 115200 and finally  to 57600?

puddly:
This is precisely what `ControllerApplication.probe` does, these settings are specific to a radio library.

MattWestb:
> The only stick that I have is the HUSBZB-1 with the original firmware (57600). All of the updated firmwares I think switch it to 115200?

Garry was cooking all community firmware for EM3X devices but i think it was both off some user was liking having the old slow HW and some like faster with SW but i think is nearly zero left of them now after Snoff and SC and so on.
If i have understanding right is only probing auto conf so if adding one coordinator manually its still possible setting the baud rate so shall working like before ding the probing = the EM3X time.

puddly:
We'll drop 57600 next release.

Issue #662, opened 2025-01-10
EZSP v16 (EZSP PROTOCOL VERSION 16) support in bellows?


Hedda:
bellows support for EZSP protocol version incremented to EZSP v16 in Zigbee EmberZNet v8.1.0.0 (from Simplicity SDK 2024.12.0):

https://www.silabs.com/documents/public/release-notes/emberznet-release-notes-8.1.0.0.pdf

Note! EZSP protocol version went from 0x0e (14) to 0x10 (16) with this EmberZNet v8.1.0.0 release, seems they jumped over v15?

Nerivec has experimental firmware builds ("_release appears mostly centered around firmware bug fixing, very few protocol changes._"):

https://github.com/Nerivec/silabs-firmware-builder/actions/runs/12538101751

_Known issues:_

Sonoff ZBDongle-E dongle is not working with [Nerivec](https://github.com/Nerivec)'s firmware build of zigbee_ncp firmware according to [Nerivec](https://github.com/Nerivec) himself. See:

* https://github.com/Nerivec/silabs-firmware-builder/discussions/29

**UPDATE!** FYI; [Nerivec](https://github.com/Nerivec) has now informed beta-test users of the Zigbee2MQTT Edge version (which tracks the dev branch of Zigbee2MQTT) that the EmberZNet 8.1.0 firmware looks to be buggy with [zigbee-herdsman](https://github.com/Koenkk/), where specifically software flow control does not seem to work properly when using EmberZNet 8.1.0 firmware with Zigbee2MQTT, (so their current recommendation for Zigbee2MQTT is to stick with 8.0.2 and newer 8.0.x releases for those who want to go with any EmberZNet 8.x.x version). He has posted more information in a bug-report to upstream here:

* https://github.com/SiliconLabs/simplicity_sdk/issues/29

darkxst:
> software flow control does not seem to work properly

This seems to have been resolved in 8.2.0 finally. This release is using EZSP v17.

Hedda:
> > software flow control does not seem to work properly
> 
> This seems to have been resolved in 8.2.0 finally. This release is using EZSP v17.

Submitted a new issue to track EZSP v17 as a separate request -> https://github.com/zigpy/bellows/issues/685

Hedda:
Missed it but looks like EZSP v16 and  EZSP v17 was added in https://github.com/zigpy/bellows/pull/678 (while EZSP v15 was skipped as there is no such version of EZSP):
 
[...]

PR #678, opened 2025-07-15
Formally support EZSP v16 and v17

puddly:
As with all protocol version changes, bellows will successfully use the most recent protocol schema when talking to an adapter: we don't need to explicitly support every protocol version bump unless fundamental commands change. There were none here.

EZSP v15 has been skipped because Simplicity SDK jumped from v14 to v16, there seemingly was no v15. There are no real changes in these releases beyond a few new command definitions that we do not use.


Hedda:
Closes https://github.com/zigpy/bellows/issues/662 and https://github.com/zigpy/bellows/issues/685 or?

* https://github.com/zigpy/bellows/issues/662

* https://github.com/zigpy/bellows/issues/685

PR #689, opened 2025-10-06
Send three resets when connecting, not just one

puddly:
Some devices, like the ZBT-2, send a single NAK if you connect to them immediately after the firmware starts. Sending more than one reset frame seems to flush enough buffers to "break through" and properly reset the device on the first try.

### DIFF
--- a/bellows/ash.py
+++ b/bellows/ash.py
@@ -715,4 +715,8 @@ class AshProtocol(asyncio.Protocol):
         )
 
     def send_reset(self) -> None:
+        # Some adapters seem to send a NAK immediately but still process the reset frame
+        # if one eventually makes it through
+        self._write_frame(RstFrame(), prefix=(Reserved.CANCEL,))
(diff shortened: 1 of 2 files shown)


TheJulianJES:
> Some devices, like the ZBT-2, send a single NAK. [...] Sending more than one reset frame seems to flush enough buffers to "break through"

Do we know why this is apparently specific to ZBT-2? Just the ESP32 USB/serial bridge being a bit slower upon startup?

TheJulianJES (approved):
In before resetting during the start of a reset breaks some coordinators with ancient firmware versions 😄 

Should be good

puddly:
I've tested all of the adapters I have, including a few newer dual protocol sticks and the *old* HUSBZB-1, both with the 115200 and the 57600 firmwares. The triple reset works on every stick. Some other adapters that required two attempts with the old behavior now work on the first try so I think this is about as well-tested as I can manage 😄.

PR #690, opened 2025-10-14
Backup/restore routing table and allow specifying TX power during formation

puddly:
Implements https://github.com/zigpy/zigpy/pull/1686 and https://github.com/zigpy/zigpy/pull/1687. Requires https://github.com/NabuCasa/silabs-firmware-builder/pull/118.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -14,10 +14,15 @@ import zigpy.zdo.types as zdo_t
 
 from bellows.ash import NcpFailure
 import bellows.config as config
-from bellows.exception import ControllerError, EzspError
+from bellows.exception import ControllerError, EzspError, InvalidCommandError
 import bellows.ezsp as ezsp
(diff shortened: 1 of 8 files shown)


tests/test_application.py
```diff
@@ -1944,6 +2032,7 @@ def zigpy_backup() -> zigpy.backups.NetworkBackup:
             channel=t.uint8_t(25),
             channel_mask=t.Channels.ALL_CHANNELS,
             security_level=t.uint8_t(1),
+            tx_power=0,
```
TheJulianJES, line 2035: For the `zigpy_backup` fixture:
Shouldn't matter for the test, but was this intentionally set to `0`, instead of e.g. `8`?

puddly: I think I set it to `0` to implicitly test `if tx_power:` type logic but that should probably have instead been made explicitly into a new test.

PR #694, opened 2025-10-24
Allow any reset during startup, not just `RESET_SOFTWARE`

puddly:
- After firmware flashing, the reset reason is usually `RESET_POWER_ON`
- After crashing due to ACK timeouts, it's `RESET_WATCHDOG`

Both of these indicate that the firmware has fully reset, we don't actually need to care about _why_ it reset and our current restriction to accepting only `RESET_SOFTWARE` as a valid reason causes bellows to fail to start up in cases where it can continue without any issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     with:
       CODE_FOLDER: bellows
       CACHE_VERSION: 2
-      PYTHON_VERSION_DEFAULT: 3.11.0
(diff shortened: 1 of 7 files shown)


TheJulianJES (approved):

davet2001:
I've tried this performing 80 upgrades in a loop using a script that previously reproduced the occasional failure (previously 3-10 failures when tried 80times).

With this PR, I got 0 failures on 80 retries.

PR #695, opened 2025-10-30
Fix writing network settings with missing TX power

puddly:
EZSP is the only radio type that expects an explicit TX power to be supplied during network formation. We really should migrate all of these radio library methods to underscore-prefixed variants (e.g. `write_network_info` to `_write_network_info`) so that zigpy can fix settings up before having the radio library perform the operation.


TheJulianJES (approved):
Looks good

PR #699, opened 2025-11-25
Fix ASH reject condition amplification bug

puddly:
When communicating with OpenThread firmware, we get into tight loops when the firmware manages to respond with a semi-plausible frame. Our ASH implementation rejects it by sending a `CANCEL` byte and a `NakFrame`, which prompts the firmware to send more data, etc. This isn't right. We should enter a "reject" state and wait until the firmware sends something sensible.

See https://github.com/home-assistant/core/issues/156966#issuecomment-3577469689.

```python
2025-11-25 21:37:31 emscripten bellows.ash[42] DEBUG Sending frame 32*CANCEL RstFrame() FLAG
2025-11-25 21:37:31 emscripten bellows.ash[42] DEBUG Sending data  1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1ac038bc7e
2025-11-25 21:37:31 emscripten bellows.ash[42] DEBUG Received data 7e8006704672616d696e67206572726f7220363a205bcafd7e7e80067031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412031412043302033382042435d0a89257e
2025-11-25 21:37:31 emscripten bellows.ash[42] DEBUG Failed to parse frame bytearray(b'\x80\x06pFraming error 6: [\xca\xfd')
Traceback (most recent call last):
  File "/lib/python3.13/site-packages/bellows/ash.py", line 470, in data_received
    frame = parse_frame(data)
  File "/lib/python3.13/site-packages/bellows/ash.py", line 350, in parse_frame
    return frame.from_bytes(data)
           ~~~~~~~~~~~~~~~~^^^^^^
  File "/lib/python3.13/site-packages/bellows/ash.py", line 219, in from_bytes
    control, data = cls._unwrap(data)
                    ~~~~~~~~~~~^^^^^^
  File "/lib/python3.13/site-packages/bellows/ash.py", line 156, in _unwrap
    raise ParsingError(
    ...<2 lines>...
    )
bellows.ash.ParsingError: Invalid CRC bytes in frame bytearray(b'\x80\x06pFraming error 6: [\xca\xfd'): expected 6e1e, got cafd
2025-11-25 21:37:31 emscripten bellows.ash[42] DEBUG Sending frame CANCEL NakFrame(res=0, ncp_ready=0, ack_num=0) FLAG
[...]

PR #700, opened 2025-11-25
Compare to enums with `is` in hot paths

puddly:
https://github.com/zigpy/bellows/pull/699#discussion_r2561582011

### DIFF
--- a/bellows/ash.py
+++ b/bellows/ash.py
@@ -396,7 +396,7 @@ class AshProtocol(asyncio.Protocol):
 
     def _reject_frame(self) -> None:
         """Send NAK and enter reject condition if not already rejected."""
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
😄

TheJulianJES:
Well, I guess these are called somewhat often, so it's not completely useless.

puddly:
It's about 20% faster but still on the order of a few nanoseconds 🤷‍♂️ 
```python
In [4]: %timeit value == NcpState.FAILED
28.7 ns ± 0.577 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit value is NcpState.FAILED
23.9 ns ± 0.232 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

PR #705, opened 2026-01-01
Fix fallback logic when starting up with an unsupported EZSP version

puddly:
I tested EZSPv18 a bit more thoroughly and while the frame format has indeed changed, it didn't actually change in any way that impacts bellows operation: it should have been supported by "falling back" to EZSPv17 parsing and logic.

ZHA startup fails because config writing tried to look up config for the exact version in our global config, which will fail. We should instead fall back to the latest version's config.

Should resolve https://github.com/home-assistant/core/issues/160051

### DIFF
--- a/tests/test_ezsp.py
+++ b/tests/test_ezsp.py
@@ -951,6 +951,32 @@ async def test_cfg_initialize_skip():
         )
 
 
+@pytest.mark.parametrize("unsupported_version", [15, 18, 99])
+async def test_unsupported_ezsp_version_startup(unsupported_version: int, caplog):
+    """Test that startup works with an unsupported EZSP version."""
+    ezsp = make_ezsp(version=unsupported_version)
+
+    with patch("bellows.uart.connect"):
+        await ezsp.connect()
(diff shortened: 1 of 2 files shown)


tests/test_ezsp.py
```diff
@@ -951,6 +951,32 @@ async def test_cfg_initialize_skip():
...
+    ezsp = make_ezsp(version=unsupported_version)
+
+    with patch("bellows.uart.connect"):
+        await ezsp.connect()
+
+    # The EZSP version should be stored as the unsupported version
+    assert ezsp._ezsp_version == unsupported_version
+
+    # But the protocol should fall back to the latest
+    assert ezsp._protocol.VERSION == EZSP_LATEST
```
TheJulianJES, line 966: EZSP v15 falling back to whatever is the latest version feels a bit weird, but should be a non-issue as that version doesn't actually exist 😄  (and that's nothing new from this PR)

TheJulianJES (approved):
nice!

Issue #110, opened 2018-04-05
Support for ZigBee "Green Power" devices such as the Hue Tap Switch


kiall:
Hey - So, I could be missing something, but as near as I can tell ZigBee "Green Power" devices, like the Philips Hue Tap Switch, are current unsupported? So far, I've tried a few things to pair a Hue Tap Switch with bellows, but have not had much luck.

Are green power devices supported? If so, is there anything special needed to pair?

mingqian:
Per my understanding, currently emberznet has very limited support for zgp functionalities, especially sink devices. For EZSP protocol, I don't think currently it supports.

Shulyaka:
Waiting for the EnOcean PTM 216Z support as well... Those battery-free devices are growing popularity now.

Adminiuga:
Where can you get one?

Adminiuga:
I think later emberznet versions had some support for ZGP, but most popular HUSBZB-1 stick is still on 4.0 version and unlikely to be updated. 
But I see that Conbee added ZGP endpoint so my guess they are gonna support it at some point.

Hedda:
FYI bellows and zigpy doesn't yet support ZGP (Zigbee Green Power) but @zoic21 started working on it now, see https://github.com/zigpy/zigpy/pull/519

Hedda:
UPDATE: Developers should see any new pull requests and follow the related zigpy development discussion in this feature request: 

* https://github.com/zigpy/zigpy/issues/341

End-users (non-developers) can discuss further in this Home Assistant's community forum thread:

* https://community.home-assistant.io/t/zha-integration-support-for-zgp-zigbee-green-power-batteryless-devices-via-zigpy-to-work-with-philips-hue-click-philips-hue-tap-and-other-friends-of-hue-switches/277149

As mentioned, Zigbee Green Power (ZGP) self-powered devices are not yet supported in the ZHA integration because they are not yet support in the zigpy library that it depends on, so you need to wait until the pull request by konistehrad that will add ZGP (Zigbee Green Power) support has been merged into zigpy and a release of it in turn gets bumped in Home Assistant core.

Issue #712, opened 2026-04-23
`gpepIncomingMessageHandler` fails to parse GPDFs from Friends of Hue switches (EZSP v ? - ZBT-1)


nmingam:
## TL;DR

A community tester running the Green Power branch of zigpy (zigpy/zigpy#1814) on a Silabs ZBT-1 stick has a Busch-Jaeger 6716 U "Friends of Hue" switch send a valid GP Commissioning frame. The NCP forwards it to the host through `gpepIncomingMessageHandler` (frame ID `0x00C5`), but bellows drops the frame with `ValueError: Data is too short`, so zigpy never sees it.

The zigpy side of the work is ready and tested against the captured payload. This issue is the blocker on the bellows side.

Related prior reports, both about the same callback:
- #110 (2018) initial GP support request.
- #229 same "not a valid EmberStatus" pattern with a Niko FoH switch.

## The failing log

Reported by @dmatscheko in zigpy/zigpy#1814
([comment](https://github.com/zigpy/zigpy/pull/1814)):

```
2026-04-22 20:22:30.643 WARNING (MainThread) [bellows.ezsp.protocol]
Failed to parse frame gpepIncomingMessageHandler:
b'7ccdec0086f8710186f871010000000000ffffffffe0ffffffffff2e02c5f21ce9ae2f9e4f85f15de37c1ccbd943870013911aec1d00000411011121314151617226062636465666768'
Traceback (most recent call last):
  File ".../bellows/ezsp/protocol.py", line 164, in __call__
    result, data = t.deserialize_dict(data, rx_schema)
  File ".../bellows/types/__init__.py", line 9, in deserialize_dict
    value, data = type_.deserialize(data)
  File ".../zigpy/types/basic.py", line 927, in deserialize
    raise ValueError("Data is too short")
2026-04-22 20:22:30.648 WARNING (MainThread) [bellows.ezsp] Failed to parse frame, ignoring
```

Environment:
- bellows: **0.49.0** (`zha==1.1.2` in Home Assistant 2026.4.2)
- zigpy: 1.2.2 (same pin) + PR #1814 branch for the GP sink code.
- Adapter: Silabs ZBT-1 (EZSP v?? firmware, exact version to be confirmed by the tester).
- Device: Busch-Jaeger 6716 U Friends of Hue switch (battery-less, source ID `0x0171F886`).

## The frame is actually well-formed

[...]

Issue #229, opened 2020-04-02
Error pairing friend of hue switch


ffavero:
I was trying to pair the battery-less Niko friends of hue switch with ZHA (home assistant 0.107.7, python 3.7.) and I got this exception, any time as soon as I tap a button of the switch:

```
2020-04-02 23:49:23 DEBUG (bellows.thread_0) [bellows.uart] Data frame: b'3122b16d2bcc94b273a63a248067e2480b4c23abed4f6e8bfdd69c760381c0a7ebcddeca7effe7c08b2d618c4623a9eb763ba5ed71a6927341d2987e'
2020-04-02 23:49:23 DEBUG (bellows.thread_0) [bellows.uart] Sending: b'8430fc7e'
2020-04-02 23:49:23 DEBUG (MainThread) [bellows.ezsp] Application frame 197 (gpepIncomingMessageHandler) received: b'7fe681002a3270012a32700197020400008109000010ffffffffff00000000a5f100201b5eff080000000007000000070424d3eb0d'
2020-04-02 23:49:23 ERROR (MainThread) [homeassistant.core] Error doing job: Exception in callback ThreadsafeProxy.__getattr__.<locals>.func_wrapper.<locals>.check_result_wrapper() at /usr/local/lib/python3.7/site-packages/bellows/thread.py:99
ValueError: 127 is not a valid EmberStatus

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.7/site-packages/bellows/thread.py", line 100, in check_result_wrapper
    result = call()
  File "/usr/local/lib/python3.7/site-packages/bellows/ezsp.py", line 221, in frame_received
    result, data = t.deserialize(data, schema)
  File "/usr/local/lib/python3.7/site-packages/bellows/types/__init__.py", line 9, in deserialize
    value, data = type_.deserialize(data)
  File "/usr/local/lib/python3.7/site-packages/bellows/types/basic.py", line 10, in deserialize
    r = cls(int.from_bytes(data[: cls._size], "little", signed=cls._signed))
  File "/usr/local/lib/python3.7/enum.py", line 310, in __call__
    return cls.__new__(cls, value)
  File "/usr/local/lib/python3.7/enum.py", line 564, in __new__
    raise exc
  File "/usr/local/lib/python3.7/enum.py", line 548, in __new__
[...]

PR #713, opened 2026-04-23
Fix and forward gpepIncomingMessageHandler for Green Power support

nmingam:
## Summary

Supports zigpy/zigpy#1814 (Zigbee Green Power in zigpy). Real GP frames from a Friends of Hue switch (Busch-Jaeger 6716 U on a ZBT-1 stick, reported by @dmatscheko) never reached the host today for two reasons:

1. The `gpepIncomingMessageHandler` callback schema, inherited from v4 up to v16, did not match the wire format. `EmberGpAddress` in particular was declared as 14 bytes of scattered fields instead of the real 10-byte `applicationId + id[8] + endpoint`, and the strict `sl_GpStatus` enum rejected the higher status bytes current firmware returns for unmatched frames (0x7C observed in the capture).
2. Even if it parsed, bellows dispatched nothing for this callback so it never reached zigpy's GreenPower manager.

### Reference Issue
https://github.com/zigpy/bellows/issues/712
### Should also resolve
https://github.com/zigpy/bellows/issues/110
https://github.com/zigpy/bellows/issues/229

## What this PR does

**Fix the callback parsing on EZSP v13 / v14 / v16**

- Rewrite `EmberGpAddress`: 10-byte layout with `.source_id` / `.gpd_ieee_address` helpers.
- Override `gpepIncomingMessageHandler` on v13 with `status: uint8_t` so unexpected bytes reach the host; v14 picks it up through `_REPLACEMENTS`.
- v16 gets its own override that appends the `SlRxPacketInfo` trailer (zigbee-herdsman gates the read on `version >= 0x10`).
- v17 / v18 keep their own schema, which now also benefits from the struct fix.

**Forward the decoded callback to zigpy**

- Add a `gpepIncomingMessageHandler` branch in `ezsp_callback_handler` that reencapsulates the flattened GPDF as a ZCL GP Notification on endpoint 242 / cluster 0x0021, mirroring the ember adapter in zigbee-herdsman.
- On v16+ the trailing `SlRxPacketInfo` supplies the proxy NWK, LQI and RSSI. On v13 / v14 the coordinator short address is used as a fallback.
- IEEE-addressed GPDs are dropped with a debug log (zigpy's GP stack does not support them yet).
- Short / malformed args are dropped silently so a firmware mismatch cannot crash the dispatcher.
- New `green_power_rx` counter to track how many GPDFs the NCP has handed us.

## Tests

- Schema-layer tests driven by the exact hex bytes @dmatscheko captured.
- v14 / v16 inheritance tests (v14 via `_REPLACEMENTS`, v16 with the trailer appended).
- Forwarding tests covering v13 / v14 shape, v16 shape, operational data command, IEEE drop, short-args drop.
- Cross-repo integration tests that restore the real `packet_received` and let the frame actually reach zigpy's `GreenPowerManager`. Two paths covered: Toggle on a pre-registered device fires `gp_command_received`; NoSecurity commissioning registers the device via `gp_device_joined` and schedules a GP Pairing.

All tests pass locally (402 / 402, excluding 3 unrelated `test_ash` timing failures pre-existing on `dev`).

## Caveats

- Depends on an unreleased zigpy (zigpy/zigpy#1814). Keeping this PR as a draft until the zigpy side stabilizes, so both libraries can ship together.
- No version bump yet for the same reason.

## Test plan

- [x] `pytest tests/` green locally
- [x] `ruff check` no new errors
- [ ] Field test with a Busch-Jaeger 6716 U on HA 2026.4 (in progress with @dmatscheko)


Hedda:
> * Depends on an unreleased zigpy ([Add Zigbee Green Power (ZGP) protocol support zigpy#1814](https://github.com/zigpy/zigpy/pull/1814)). Keeping this PR as a draft until the zigpy side stabilizes, so both libraries can ship together.

If this PR does not break stuff if merged before https://github.com/zigpy/zigpy/pull/1814 is merged into zigpy then suggest merge https://github.com/zigpy/bellows/pull/713 first. Or? As far as I understand the normal order is to get upstream patches merge first, and technically bellows is upstream of zigpy.

That is, if nothing breaks then there should be no need for both libraries can ship together, instead updates to bellows (and other radio libraries for zigpy too) can be reviewed, merged, and released directly as long as there no breaking changes. Or?

nmingam:
Good point, and I agree.
The usual order is radio lib first, so if this PR doesn't break anything on its own, it should merge ahead of zigpy/zigpy#1814. On the breakage question: it doesn't. Once merged, bellows forwards `gpepIncomingMessageHandler` callbacks as ZCL GP Notifications on endpoint 242, cluster 0x0021. In current zigpy (without #1814), those packets go through `ControllerApplication.packet_received`, find no handler for that endpoint, and are silently dropped. Non-GP traffic is unaffected.

You're right that the type coupling was telling a different story. The previous version of this PR imported `GP_ENDPOINT` / `GP_CLUSTER_ID` from `zigpy.zgp.types`, which are only added by zigpy/zigpy#1814
I just commited an inline version, those constants are standard defined. And this way this PR does no more rely on the zigpy PR.

Hedda:
> this PR doesn't break anything on its own, it should merge ahead of https://github.com/zigpy/zigpy/pull/1814. On the breakage question: it doesn't.
> …
> this PR does no more rely on the zigpy PR.

So maybe change this from a draft PR to ready for review, unless not ready yet?

nmingam:
# ZGP pairing flow: field test report (Busch-Jaeger 6716 U on HA 2026.4)

## Context

- Setup: ZBT-1 stick, bellows fork (PR #713), zigpy fork (PR #1814)
- Test device: Busch-Jaeger 6716 U "Friends of Hue"
- Action: "Settings → Zigbee → + Add Device", then the commissioning button combo on the switch (top-left 10s hold, then top-right + bottom-left simultaneously).
- Debug loggers enabled: `bellows.ezsp`, `zigpy.zgp.manager`.

## TL;DR

The EZSP -> bellows -> zigpy forwarding path is **working end-to-end**. The blocker is one layer above: the ZGP commissioning window on the sink is never opened.
This is a missing wiring in the zigpy PR (#1814), not a bellows issue.

## What works

- `gpepIncomingMessageHandler` fires on every GPDF (cmds 0x11, 0x68, 0xE0)
- `EmberGpAddress` is decoded correctly (`source_id=0x0171F886`, `applicationId=0`, security fields intact)
- `packet_received` is called with `src_ep=242`, `dst_ep=242`, `profile_id=0xA1E0`, `cluster_id=0x0021`
- Frames are transmitted to the zigpy ZGP manager:
  - `zigpy.zgp.manager: GP Notification: source_id=0x0171F886, cmd=...` is visible for every frame
  - `zigpy.zgp.proxy: New proxy entry: GPD 0x0171F886 via proxy 0x0000` (coordinator acting as its own proxy)

## To be tested

ZGP frames with a legacy zigpy version should be silently dropped. This would match what happened with recent versions of the EZSP protocol already handling ZGP frames.

bellows/types/struct.py
```diff
@@ -357,16 +357,43 @@ class EmberTokTypeStackZllSecurity(EzspStruct):
...
+    ``gpd_ieee_address`` helpers below to access the right view.
+    """
+
+    # The GPD Application ID: 0 = source ID mode, 2 = IEEE mode.
     applicationId: basic.uint8_t
+    # Raw 8-byte union. Use :attr:`source_id` or :attr:`gpd_ieee_address`
+    # to interpret it based on :attr:`applicationId`.
+    id: basic.FixedList[basic.uint8_t, 8]
     # The GPD endpoint.
     endpoint: basic.uint8_t
```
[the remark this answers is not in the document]
puddly: Not necessary.

bellows/ezsp/v13/commands.py
```diff
@@ -90,6 +90,42 @@
             "status": t.sl_Status,
         },
     ),
```
puddly, line 92: These verbose Claude comments are not useful, please remove them.

bellows/types/struct.py
```diff
@@ -357,16 +357,43 @@ class EmberTokTypeStackZllSecurity(EzspStruct):
 
 
 class EmberGpAddress(EzspStruct):
```
puddly, line 359: The SDK source shows the definition as:

```C
/**
 * @brief GPD Address for sending and receiving a GPDF.
 */
typedef struct {
  union {
    /** The IEEE address is used when the application identifier is
     *  ::SL_ZIGBEE_GP_APPLICATION_IEEE_ADDRESS.
     */
    sl_802154_long_addr_t gpdIeeeAddress;
    /** The 32-bit source identifier is used when the application identifier is
     *  ::SL_ZIGBEE_GP_APPLICATION_SOURCE_ID.
     */
    sl_zigbee_gp_source_id_t sourceId;
  } id;
  /** Application identifier of the GPD. */
  sl_zigbee_gp_application_id_t applicationId;
  /** Application endpoint , only used when application identifier is
   * ::SL_ZIGBEE_GP_APPLICATION_IEEE_ADDRESS
   */
  uint8_t endpoint;
} sl_zigbee_gp_address_t;
```

This differs from the field order in this struct.

bellows/zigbee/application.py
```diff
@@ -75,6 +78,13 @@
 
 DEFAULT_TX_POWER = 8  # dBm
 
+# Zigbee Green Power spec constants — hardcoded rather than imported from
+# ``zigpy.zgp`` so bellows does not depend on the upstream zigpy release
+# shipping these symbols. These values are fixed by the ZGP 1.1b profile.
```
puddly: This is a pretty verbose comment and should be replaced with a TODO to remove these constants, since we should use the ones from zigpy once they are released in the package.

bellows/zigbee/application.py
```diff
@@ -696,6 +706,119 @@ def ezsp_callback_handler(self, frame_name, args):
...
+        duplicate suppression, etc.) and hands us a flattened GPDF. zigpy's
+        GP manager on the other side expects a ZigbeePacket on endpoint 242,
+        cluster 0x0021, carrying a ZCL GP Notification (server command 0x00)
+        so it can route commissioning, data, and decommissioning frames
+        through the same entry point.
+
+        We therefore rebuild that envelope here, mirroring what zigbee-
+        herdsman's ember adapter does. Any GPDF that cannot be converted
+        (for example an IEEE-addressed frame, which the current zigpy GP
+        stack explicitly does not support) is logged and dropped.
```
puddly: Same here, the docstring can be removed entirely and replaced with just the first line.

bellows/zigbee/application.py
```diff
@@ -696,6 +706,119 @@ def ezsp_callback_handler(self, frame_name, args):
...
+            payload=zigpy.types.LVBytes(bytes(gpd_command_payload)),
+        )
+
+        # ZCL header: frame_control=0x01 (cluster-specific, client→server,
+        # not manufacturer-specific, no default response disabled), TSN is
+        # the GPD MAC sequence number so duplicates can be tracked, and the
+        # command is GP Notification (0x00).
+        zcl_bytes = (
+            bytes([0x01, int(sequence_number) & 0xFF, 0x00]) + notification.serialize()
+        )
```
puddly: Construct the `t.ZCLHeader` object directly to avoid magic byte arrays.

bellows/zigbee/application.py
```diff
@@ -696,6 +706,119 @@ def ezsp_callback_handler(self, frame_name, args):
...
+
+        if packet_info is not None:
+            proxy_nwk = int(packet_info.sender_short_id)
+            lqi = int(packet_info.last_hop_lqi)
+            rssi = int(packet_info.last_hop_rssi)
+        else:
+            proxy_nwk = int(self.state.node_info.nwk)
+            lqi = int(gpd_link)
```
puddly: Is this right?

tests/test_application.py
```diff
@@ -541,6 +541,155 @@ def test_frame_handler_ignored(app, aps_frame):
...
+    ]
+
+
+def test_gp_frame_handler_forwards_notification(app):
+    """A GPDF is forwarded to zigpy as a ZCL GP Notification."""
+    app.ezsp_callback_handler("gpepIncomingMessageHandler", _gp_args_v13())
+
+    assert app.packet_received.call_count == 1
+
+    packet = app.packet_received.mock_calls[0].args[0]
```
puddly: Compare it with `==` to the expected packet object

bellows/types/struct.py
nmingam: I tried matching the SDK order but the captured frame from @dmatscheko test fails to parse: the EZSP wire layout serializes applicationId first, even though the C struct declares id first.

bellows/zigbee/application.py
nmingam: Replaced the paragraph with a TODO in 2df0ea5

bellows/zigbee/application.py
nmingam: Resumed in a single line doscstring in 2df0ea5

bellows/zigbee/application.py
nmingam: Done in 2df0ea5

bellows/zigbee/application.py
nmingam: This is the v13/v14 fallback when no SlRxPacketInfo trailer is present.
Would you rather drop a synthetic LQI/RSSI on v13/v14 (e.g. lqi=0, rssi=0) ?

nmingam: Comment added in 2df0ea5

tests/test_application.py
nmingam: Done in af28f20

bellows/zigbee/application.py
```diff
@@ -696,6 +706,119 @@ def ezsp_callback_handler(self, frame_name, args):
...
+
+        # v16+ appends an SlRxPacketInfo after the payload. Pull the
+        # proxy NWK out of it if it is there, otherwise fall back to the
+        # coordinator short address (the coordinator acts as the proxy).
+        packet_info = rest[0] if rest else None
+
+        if addr.applicationId == zgp_t.ApplicationID.SrcID:
+            source_id = addr.source_id
+        else:
+            LOGGER.debug(
```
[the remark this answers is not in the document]
nmingam: Taken into account in 2df0ea5
Added isinstance(addr, t.EmberGpAddress) guard with a debug log and a regression test

bellows/ezsp/v16/commands.py
```diff
@@ -1,5 +1,33 @@
...
 from ..v14.commands import COMMANDS as COMMANDS_v14
 
 COMMANDS = {
     **COMMANDS_v14,
+    # EZSP v16 appends an ``EmberRxPacketInfo`` struct after the LVBytes
+    # payload in ``gpepIncomingMessageHandler``. zigbee-herdsman gates the
+    # read on ``version >= 0x10`` (see
+    # ``src/adapter/ember/ezsp/ezsp.ts`` in the ``ember`` adapter), so v13
+    # and v14 do not carry this trailer but v16 does. v17 and v18 override
+    # the whole command again, so this entry only matters for v16.
```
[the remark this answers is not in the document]
nmingam: Done in 4bc2d50

bellows/ezsp/v16/commands.py
```diff
@@ -1,5 +1,28 @@
...
+            "gpdfSecurityLevel": t.EmberGpSecurityLevel,
+            "gpdfSecurityKeyType": t.EmberGpKeyType,
+            "autoCommissioning": t.Bool,
+            "bidirectionalInfo": t.uint8_t,
+            "gpdSecurityFrameCounter": t.uint32_t,
+            "gpdCommandId": t.uint8_t,
+            "mic": t.uint32_t,
+            "proxyTableIndex": t.uint8_t,
+            "gpdCommandPayload": t.LVBytes,
+            "packetInfo": t.SlRxPacketInfo,
```
puddly: I've tested this with an adapter running EZSPv13 and EZSPv18 and parsing fails because `packetInfo` is not actually part of the command payload.

nmingam:
Status update.

Dropped the `SlRxPacketInfo` trailer @puddly flagged: it isn't in the wire format on v13 or v18 (you tested), and the v17 override that came in via upstream #678 had the same problem. v13/v14/v16/v17/v18 all share one `gpepIncomingMessageHandler` schema now, no trailer. Simplified `_handle_gp_frame` accordingly (no more conditional `packet_info` branch).

On the merge-order question @Hedda raised earlier: this PR doesn't depend on zigpy/zigpy#1814 anymore. Once merged, GP frames forward through `packet_received` and get silently dropped by current zigpy (no handler for endpoint 242). Non-GP traffic is unaffected.

Hedda:
Any updates on this?

nmingam:
Sorry, I've been short on time to work on this recently. I'll probably be a bit more available by the end of June to fully validate it, even though the tests performed by @dmatscheko on https://github.com/zigpy/zigpy/pull/1814 seem pretty encouraging.

nmingam:
@mineshaftgap mentioned a v4 `gpepIncomingMessageHandler` schema fix this PR is missing. Happy to take it here, either as a PR against this branch or a patch.

mineshaftgap:
Thanks - opened as a PR against this branch: nmingam/bellows#1.

It's the v4 base schema (inherited by v4..v12), which still has the five scattered address fields and
`bidirectionalInfo: Bool`. The fix swaps those for the `EmberGpAddress` struct and
`bidirectionalInfo: uint8_t`, matching your v13+ fix. v4 keeps `status: EmberStatus` (v13+ moved to
`uint8_t`). Second commit is a v4 test that parses your existing `BJ6716U_GPEP_PAYLOAD` capture against
the v4 schema end-to-end, so every EZSP version is now exercised against the same real bytes - it needs
nothing this branch doesn't already have.

mineshaftgap:
Heads up - the red CI here isn't from the v4 merge; the run on the prior HEAD (`27e606f`) was already
red. Two separate causes, both outside the GP code:

**1. pre-commit / `application.py` formatting.** The pinned `black` (23.3.0) wants to reformat the
`hashed_tclk` assignment in `bellows/zigbee/application.py`, so `pre-commit run --all-files` flags it:

```diff
@@ -543,13 +543,13 @@
         if use_hashed_tclk and not stack_specific.get("hashed_tclk"):
             # Generate a random default
-            network_info.stack_specific.setdefault("ezsp", {})["hashed_tclk"] = (
-                os.urandom(16).hex()
-            )
+            network_info.stack_specific.setdefault("ezsp", {})[
+                "hashed_tclk"
+            ] = os.urandom(16).hex()
```

The file looks formatted by a newer `black` (24+) than the pinned 23.3.0, so the two are fighting -
running the pinned pre-commit reverts it to 23.3.0 style and makes the check green, or bumping the pin
keeps the current style. (The GP/v4 files are already clean against 23.3.0.)

**2. tests on 3.11 + 3.13.** The only failure is an ERROR at *teardown* of
`tests/test_thread.py::test_thread_already_stopped` ("RuntimeError: Event loop is closed"). It passes on
3.12 and 3.14 in the same run and 12/12 locally for me, so it looks like a flaky event-loop teardown
rather than a real regression - no GP code involved.

Happy to send a small PR with just that `application.py` formatting hunk if that's useful.

bellows/types/struct.py
nmingam: Following up on this with the byte-level evidence, since it's the one open question left on this PR.

The captured `gpepIncomingMessageHandler` payload from a real ZBT-1 (Busch-Jaeger 6716U, pinned in `tests/test_ezsp_v13.py`) starts:

```
7ccdec 00 86f8710186f87101 00 ...
│      │  │                └─ endpoint
│      │  └─ id (8-byte union)
│      └─ applicationId
└─ status, gpdLink, sequenceNumber
```

Parsing the 10 address bytes `0086f8710186f8710100` both ways:

| Field order | SrcID decoded | applicationId |
| --- | --- | --- |
| `applicationId` first (this PR) | `0x0171F886` ✅ | `0` = SrcID ✅ |
| SDK declaration order (`id` first) | `0x71F88600` ❌ | `1` = LPED ❌ |

The device's actual source ID is `0x0171F886` and it commissions in SrcID mode, so only the first ordering decodes correctly — the second shifts everything by one byte and yields a nonexistent device in the wrong application mode.

So the C struct's declaration order and the EZSP wire order genuinely differ here; `sl_zigbee_gp_address_t` is not serialized field-for-field over EZSP. The struct carries a docstring explaining this and the real-frame test pins it. Happy to reshape how the layout is expressed if you'd prefer, but the byte order itself has to stay as-is or parsing breaks.

PR #714, opened 2026-04-28
Increase startup reset wait to 2 seconds

agners:
When connecting over a TCP serial socket (e.g. the silabs-multiprotocol add-on), the NCP reset frame can take ~1.5s to arrive after the remote reuses the socket from a previous instance. The 1s wait causes a spurious retry on every startup; bumping to 2s lets the reset land within the first attempt.


agners:
I've also noticed that you've experimented with 2s as well in #485 (see https://github.com/zigpy/bellows/pull/485/changes/f2b961652c6b243daf80af7132dd09b35fe04cc9). It does work with 1s every now and then (maybe 1 out of 10 attempts), but 2s makes it really reliable.

PR #718, opened 2026-05-04
Align MTORR concentrator constants with Silicon Labs SDK defaults

TheJulianJES:
DRAFT: Check if this is really accurate.

## AI summary (check if accurate)

The current values were inherited from very old EmberZNet sample code and don't match what the Silicon Labs SDK or other modern adapters use:

| Param                       | Old   | New | SDK default | SDK range |
|-----------------------------|-------|-----|-------------|-----------|
| MTOR_MIN_INTERVAL           | 60    | 10  | 10          | 1..60     |
| MTOR_MAX_INTERVAL           | 3600  | 60  | 60          | 30..300   |
| MTOR_ROUTE_ERROR_THRESHOLD  | 8     | 3   | 3           | 1..100    |
| MTOR_DELIVERY_FAIL_THRESHOLD| 8     | 1   | 1           | 1..100    |

`MTOR_MAX_INTERVAL = 3600` was the worst of the four — 12× the SDK documented maximum. The NCP accepted it on the wire (uint16) but it meant up to an hour between MTORR broadcasts, leaving devices that lost their route to the concentrator stranded for that long. The higher error/failure thresholds also slowed reactive re-broadcasts on topology changes.

The SDK plugin defaults are documented in
`protocol/zigbee/app/framework/plugin/concentrator/config/concentrator-config.h`. zigbee-herdsman's ember adapter applies the same defaults and validates against the same ranges.


TheJulianJES:
I think this is something we should include. The current values are broken (one outside of allowed range) and the new ones also match Z2M, so they can't be too bad.

At some point in the future, I want to have a closer look at this though, as my coordinator sometimes uses very outdated source routes when sending requests, even though the routers changed network addresses days ago, so the messages just never arrive. I do wonder if this is related to us incorrectly setting up source routing... somehow?

puddly (approved):

PR #728, opened 2026-07-07
Set default CONFIG_BROADCAST_TABLE_SIZE

MattWestb:
Zigbee PRO must have it set to 15 and if user like doing it different they can setting it in config on there own risk braking broadcast in the mesh network (if using star network its not a problem)


MattWestb:
Sorry @puddly for pinging you in the morning **!!!**
I have putting in default for broadcast table but one test is not likeing it but i think its unrelated to my change.

This is the right way if like hotting the broadcast with "extended firmware" it shall being done in config like:
```
    ezsp_config:
      CONFIG_MAX_END_DEVICE_CHILDREN: 0
      CONFIG_BROADCAST_TABLE_SIZE: 254
```
I running my IKEA controller test network with the config and its with the stock ZBT-2 firmware.
My production system is running on Billy EZSP and have Silabs standard so shall being 15 in the firmware.

puddly:
@MattWestb Given https://github.com/zigpy/ziggurat/issues/43 and https://github.com/zigpy/ziggurat/pull/55, do you still think SiLabs default of 15 is right? Nobody really uses such a conservative config even for manufacturers using SiLabs chips, no?

MattWestb:
Its up to you but its depends of the users network if all routers is "extended" is it OK.
I think the best / safe way for not getting problem is making it default 15 and have instruction in the wiki for overriding it (if the firmware is supporting it) then users is (hopefully) understand what was going wrong with there adaptive lighting in broadcast mode.
My opinion is ZHA shall being default "safe configured" out of the box and if like tuning it then the user is doing it on own risk and not the opposite (Z2M is allowing have joining open 24/7 and most users is not knowing it and what it can happening).

Feel free do what you like / feel is the right for ZHA in the end = fixing or closing.

## As normal great work / digging done !!

TheJulianJES:
While I totally understand the concern behind the varying broadcast limits, I think the community feedback so far was that the higher broadcast table size actually helped them.

In most cases, I think it's only really relevant for users manually dragging the color wheel slider in the HA UI. And there, you either get an error message really fast with the `15` limit or everything works for much longer with the higher limits. In the worst case, users may see some (further away) routers not respond with the higher limit, but with the [increased broadcast limits that many manufacturers](https://github.com/zigpy/ziggurat/pull/55#issue-5003946209) (of routers) seem to use with SiLabs, I think that's not a huge concern.

Should this turn out to be an actual issue, we could even add a settings toggle in the frontend which is like "Limit broadcast messages" with an explanation to use that when you have a busy network or congested airtime, perhaps (to reserve "airtime" for unicasts, even if a lot of broadcast commands are being transmitted).

### Grouped Hue lights testing

I do have 11 Hue lights grouped in a garden I've just tested with and even spamming broadcast commands (dragging color wheel sends commands every 500 ms), everything seems to react just fine there. They're all not "directly connected" to the coordinator, most have Hue lights in between.

Only when spamming them really hard after some time (rapidly clicking color wheel as fast as possible, so broadcast commands get sent every few millisecond) do I see `sl_Status.ZIGBEE_MAX_MESSAGE_LIMIT_REACHED` errors (which are actually `NETWORK_BUSY` but [the mapping](https://github.com/zigpy/bellows/blob/553ec35059c5062cb3e82e9879cbaf9eb0a373ab/bellows/types/named.py#L1716) for older Gecko SDK-based firmwares currently changes it to `ZIGBEE_MAX_MESSAGE_LIMIT_REACHED` displayed in the frontend's error message, which is a bug I'll put up a PR for).

So at that point, I do manage to hit the broadcast table size limit again. It also seems like some lights rarely do desynchronize when I really spam it. But I guess that's because the 64 limit is higher than the one Hue is using*. IMO, it's not a big issue though, as you have to hit it really hard to cause that. I think this is acceptable.

(*: Limit for outgoing broadcasts would be 58 – the bot comment will have more on this. But even that number is likely still slightly higher than the Hue limit, though again, I don't think it's a big issue currently.)

### Right balance

Overall, I think this boils down to figuring out a good balance. That manufacturers already use higher broadcast limits than the default `15`, even though it is warned against, is because the experience with Zigbee (light) groups is just somewhat bad otherwise. So I think we should also keep the higher limit by default to have a comparable experience to using the Hue/IKEA Bridge with light groups directly, for example. 

### On this PR

I'll have the bot write something as well here to see if there's extra information but I think the PR is not doing what you want it to do at the moment. Currently, this PR would always set a minimum of `15` for the broadcast table size. This is already enforced by the SiLabs SDK at compile time IIRC. If a firmware is compiled with a higher value, this PR would not change that.

MattWestb:
Thanks @TheJulianJES for one deep explanation.
So in short you is saying is drop `minimum=True` and it shall working !

The rest i think you and your KI have not understating all parts.
Refereeing to OT and other thing is use less if not understanding how 15.4 is working under the Zigbee network and application stack.
And the only indication we can getting then somthing is wrong outside in the network is `CSMA/CCA channel access failures` and one not working mesh network.
The clue is 802.15.4 is one ad-hock mac layer protocol and nothing else.

So pleas read and understand how `docs-06-3474-23-csg-zigbee-specificationR23.2_clean.pdf` and `22-65816-030-PRO-BDB-v3.1-Specification.pdf` is working on top of `802.15.4-2006.pdf`.

Also i was finding one little old but very deep and good paper (from TUWien ca 10 Km from me) `Bunyai Dominik - 2012 - ZigBee network layer simulation on top of IEEE 802154.pdf` from that you can doing little real life calculation with for implanting broadcast adaptive lighting in ZHA without braking one large (around 100 or more devices) network.

I attaching the no standard papers that is not easy to find on the internet for you and your KI for digging in.
15.4 version 2006: [802.15.4-2006.pdf](https://github.com/user-attachments/files/30938207/802.15.4-2006.pdf)
And deep diving in Zigbee broadcast / 15.4: 
[Bunyai Dominik - 2012 - ZigBee network layer simulation on top of IEEE 802154.pdf](https://github.com/user-attachments/files/30938218/Bunyai.Dominik.-.2012.-.ZigBee.network.layer.simulation.on.top.of.IEEE.802154.pdf).


PS: If Puddly like / need one explanation (i think hi dont) without digging thru all docks hi is welcome getting one very short explanation off line and without KI.

Issue #654, opened 2024-12-03
Controller doesnt reconnect after serial connection lost.


ryanwinter:
Previously in home assistant 2024.10.3, I would see the following errors, but everything recovered fine:
```
2024-12-03 11:58:02.028 ERROR (bellows.thread_0) [bellows.uart] Lost serial connection: SerialException('device reports readiness to read but returned no data (device disconnected or multiple access on port?)')
2024-12-03 11:58:02.029 ERROR (MainThread) [bellows.ezsp] NCP entered failed state. Requesting APP controller restart
```

But now (home assistant 2024.11), it seems that if the serial connection is lost, it doesnt appear to reestablish? I think it might be related to #633 

This issue is also being tracked at in [Home Assistant Core](https://github.com/home-assistant/core/issues/130548).

puddly:
Can you attach a full ZHA debug log of it failing to reconnect?

erkr:
> Can you attach a full ZHA debug log of it failing to reconnect?

The are several logs in the references issue https://github.com/home-assistant/core/issues/130548. 
Failed reconnect will be part of them

pipiche38:
+1 having similar issue when using zigpy + bellows fromthe Z4D plugin

puddly:
@erkr Your logs seem to specifically omit the logging I'm interested in. I need to see if this occurs with other people because reconnection logic is working as far as I'm able to test with SiLabs radios.

erkr:
@puddly pity they don't provide what you need. I was able to return my ZBDongle-E and reverted to my Conbee2. Less range but stable

ryanwinter:
I'm still running ha 2024.10.3 without any issues, I'll try and roll forward again and capture more logs but unfortunately I'm away into early Jan.

ktrask:
I have the same issue as described in the bug, that the serial connection is not reconnecting. I hope my logs are helpful. This error is most noticable: `TypeError: object NoneType can't be used in 'await' expression`

Logs:

```
Failed to send request: ApplicationController is not running


Logger: py.warnings
Quelle: /usr/local/lib/python3.13/warnings.py:110
Erstmals aufgetreten: 09:01:55 (1 Vorkommnisse)
Zuletzt protokolliert: 09:01:55

/usr/local/lib/python3.13/asyncio/base_events.py:746: RuntimeWarning: coroutine 'SerialProtocol.disconnect' was never awaited self._ready.clear()

Logger: zigpy.application
[...]

PR #730, opened 2026-07-16
Handle async proxy calls after event loop closes

Wackymax:
## What changed

- make coroutine methods exposed by `ThreadsafeProxy` consistently return an awaitable wrapper
- treat calls made after the worker event loop has closed as a disconnected no-op
- add a regression test covering an async proxy method after loop closure

## Why

When a threaded EZSP transport loses its connection, the worker event loop can already be closed by the time ZHA asks bellows to disconnect. The existing proxy logs `Attempted to use a closed event loop` and returns plain `None`, even when the proxied method is asynchronous. The caller then attempts to await that value and raises:

```text
TypeError: object NoneType can't be used in 'await' expression
```

This secondary cleanup exception can prevent Home Assistant's ZHA config-entry reload from recovering after the original transport failure.

The async wrapper now remains awaitable in every state. If the target loop is closed, awaiting it returns `None`, matching the existing disconnected/no-op behavior without raising a second exception.

Related reports:

- https://github.com/home-assistant/core/issues/130548
- https://github.com/zigpy/bellows/issues/654

## Validation

- added `test_proxy_async_loop_closed`
- `python -m pytest -q`: 425 passed
- `git diff --check`

### DIFF
--- a/bellows/ezsp/__init__.py
+++ b/bellows/ezsp/__init__.py
@@ -221,7 +221,11 @@ class EZSP:
     async def disconnect(self):
         self.stop_ezsp()
         if self._gw:
-            await self._gw.disconnect()
+            try:
+                await self._gw.disconnect()
+            except ConnectionResetError:
+                # A closed gateway loop is already disconnected.
+                pass
             self._gw = None
 
     async def _command(self, name: str, *args: Any, **kwargs: Any) -> Any:
--- a/bellows/thread.py
+++ b/bellows/thread.py
@@ -1,6 +1,7 @@
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import functools
+import inspect
 import logging
 
 LOGGER = logging.getLogger(__name__)
@@ -88,18 +89,18 @@ class ThreadsafeProxy:
                 )
             )
 
-        if asyncio.iscoroutinefunction(func):
+        if inspect.iscoroutinefunction(func):
 
             async def async_func_wrapper(*args, **kwargs):
+                """Run async proxy calls when this returned coroutine is awaited."""
                 loop = self._obj_loop
                 curr_loop = asyncio.get_running_loop()
                 call = functools.partial(func, *args, **kwargs)
                 if loop == curr_loop:
                     return await call()
                 if loop.is_closed():
-                    # Disconnected
                     LOGGER.warning("Attempted to use a closed event loop")
-                    return None
+                    raise ConnectionResetError("Attempted to use a closed event loop")
                 future = asyncio.run_coroutine_threadsafe(call(), loop)
                 return await asyncio.wrap_future(future, loop=curr_loop)
 
--- a/bellows/thread.py
+++ b/bellows/thread.py
@@ -88,6 +88,23 @@ class ThreadsafeProxy:
                 )
             )
 
+        if asyncio.iscoroutinefunction(func):
+
+            async def async_func_wrapper(*args, **kwargs):
+                loop = self._obj_loop
+                curr_loop = asyncio.get_running_loop()
+                call = functools.partial(func, *args, **kwargs)
+                if loop == curr_loop:
+                    return await call()
+                if loop.is_closed():
+                    # Disconnected
+                    LOGGER.warning("Attempted to use a closed event loop")
+                    return None
+                future = asyncio.run_coroutine_threadsafe(call(), loop)
+                return await asyncio.wrap_future(future, loop=curr_loop)
+
+            return async_func_wrapper
+
         def func_wrapper(*args, **kwargs):
             loop = self._obj_loop
             curr_loop = asyncio.get_running_loop()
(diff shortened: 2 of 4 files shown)


realKim-dotcom:
**Field report: this reproduces roughly daily on bellows 1.0.0 / HA 2026.8.2, and this PR fixes the cleanup half of it.**

@puddly asked on #720 for logs from a real installation showing the problem — here is one, on a non-SMLIGHT coordinator.

**Setup.** SONOFF Dongle Max (EFR32MG24 NCP) used as a *network* coordinator over WiFi, `socket://<lan-ip>:6638` (raw TCP bridge, single client slot). Home Assistant 2026.8.2 (Container), Python 3.14.6, bellows 1.0.0, zigpy 2.1.0, zha 2.1.0. Same site as my comment on home-assistant/core#168432 from August — it was on bellows 0.49.2 then, and the behaviour did not change across the upgrade to 1.0.0 (consistent with #727/#739 not touching the proxy).

**Frequency.** An external watchdog polls the ZHA config entry state every 5 minutes. Since 2026-08-11 the entry has been stuck in a non-`loaded` state on **14 of 26 days** (08-11, 13, 14, 15, 16, 19, 20, 21, 26, 30, 09-01, 02, 03, 05). Reloading the entry has never recovered it; a full Home Assistant restart always has.

**Latest episode, 2026-09-05** (WARNING level; three of the failure modes discussed in #722 / #740 / #745 appear in one incident):

```
16:21:22.511 WARNING (bellows.thread_0) [bellows.uart] Received an unexpected reset: <NcpResetCode.RESET_SOFTWARE: 11>
16:26:49.145 WARNING (bellows.thread_0) [py.warnings] ...: RuntimeWarning: coroutine 'Gateway.wait_for_startup_reset' was never awaited
16:26:51.140 WARNING (MainThread) [bellows.thread] Attempted to use a closed event loop   (x5)
16:26:51.142 WARNING (MainThread) [zigpy.application] Failed to disconnect from radio
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 149, in startup_reset
    await self._startup_reset()
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 131, in _startup_reset
    await self._gw.wait_for_startup_reset()
TypeError: 'NoneType' object can't be awaited

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.14/site-packages/bellows/zigbee/application.py", line 189, in connect
    await self._ezsp.connect(use_thread=self.config[CONF_USE_THREAD])
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 167, in connect
    await self.startup_reset()
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 161, in startup_reset
    await self.disconnect()
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 231, in disconnect
    await self._gw.disconnect()
TypeError: 'NoneType' object can't be awaited

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.14/site-packages/zigpy/application.py", line 329, in startup
    await self.connect()
  File "/usr/local/lib/python3.14/site-packages/bellows/zigbee/application.py", line 198, in connect
    await self._ezsp.disconnect()
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 231, in disconnect
    await self._gw.disconnect()
TypeError: 'NoneType' object can't be awaited

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.14/site-packages/zigpy/application.py", line 602, in shutdown
    await self.disconnect()
  File "/usr/local/lib/python3.14/site-packages/bellows/zigbee/application.py", line 615, in disconnect
    await self._ezsp.disconnect()
  File "/usr/local/lib/python3.14/site-packages/bellows/ezsp/__init__.py", line 231, in disconnect
    await self._gw.disconnect()
TypeError: 'NoneType' object can't be awaited
```

The identical cascade repeated at 16:30:51. The `never awaited` RuntimeWarning two seconds before it is the check-then-use race from #740/#741 (coroutine created, `run_coroutine_threadsafe` raised on the closing loop, coroutine never closed). After that the config entry went `setup_retry` → `setup_in_progress` and **stayed in `setup_in_progress` for 205 minutes** — a hang, not a crash, which matches the third row of the table in #745 (loop stopped but not yet closed → the proxied future never resolves). It cleared only with an HA restart at 20:01.

I agree with the point made on core#168432 that the *hiccup* is the coordinator's doing (the unexpected `RESET_SOFTWARE` above is the NCP's). What is bellows' doing is that the hiccup cannot be recovered from without a process restart: every teardown path — reconnect, unload, reload — dies on `ezsp/__init__.py:231` before `self._gw = None`, and `ControllerApplication.disconnect()` before `self._ezsp = None`.

**On this PR.** Before finding this thread I had patched the same three sites locally (proxy raises on a closed loop; `EZSP.disconnect()` tolerates it and always drops `_gw`). #730 applies cleanly to the 1.0.0 tag and the full test suite passes with it; the six-line reproducer (`ThreadsafeProxy(obj, loop)`, `loop.close()`, `await proxy.disconnect()`) goes from `TypeError` to the intended `ConnectionResetError`. Two small suggestions:

- `ControllerApplication.connect()`'s `except` branch and `ControllerApplication.disconnect()` (`zigbee/application.py` around lines 196–199 and 612–616) have the same shape — `await self._ezsp.disconnect()` then `self._ezsp = None` — so if `disconnect()` raises for any other reason the application keeps a dead EZSP. Moving the assignment into a `finally` closes that.
- As #745 points out, this makes teardown *complete* but does not by itself make the old transport get closed when the worker loop is torn down first, so the reconnect may still not succeed. Happy to test a patch for that ordering here — this installation reproduces the wedge often enough to give an answer within days.

**Workaround in the meantime**, if useful to anyone reading: `zha: → zigpy_config: → use_thread: false` in `configuration.yaml` (bellows' `CONF_USE_THREAD`) runs the serial link on HA's main loop, so there is no worker loop for the proxy to lose. Running that as of today; I will report back whether the wedges stop.

*(Log triage and the bench checks above were done with Claude Code; the installation, the watchdog data and the workaround run are real.)*

realKim-dotcom:
Report-back on the `use_thread: false` workaround, 11 days in (bellows 1.0.1 / HA 2026.9.1 now, same SONOFF Dongle Max over `socket://`):

- **Before** (threaded default, 2026-08-11 → 09-05): the config entry stuck non-`loaded` on 14 of 26 days, every time until a Home Assistant restart; reloading never recovered it.
- **After** (2026-09-06 13:41 → 09-17): the 5-minute poller saw 9 outages. Five were a single sample (≤10 min), one was two samples, and on the night of 09-10 the entry flapped between `setup_retry`, `setup_in_progress` and `loaded` for about four hours (03:35–07:30) — other 2.4 GHz devices on the same WiFi dropped at the same minutes, so that one is the link, not bellows. **All nine recovered on their own through HA's setup retry; zero restarts for ZHA.** The longest the entry sat in `setup_in_progress` was three consecutive samples (~15 min) on that flapping night, versus 205 min before.

So with no worker loop to lose, a lost transport during startup fails and retries the way it should. That is consistent with #751's diagnosis (the retries into a torn-down worker loop are the wedge), and I have reproduced #751 fixing it locally — numbers on that PR.

